Issue #136, opened 2018-06-25
Branding colors - coordinating with Brand team


happyimadesignr:
Conversation with Suzan (Moth and Rust) and Jay (from the Brand Working Group)

NoopDog:
![image.png](https://images.zenhubusercontent.com/57323611bf93871ccd9ad094/d26cfa51-3ade-48d7-aa2a-90d8d45c3248)

@frano-m our outcome from the call with Moth and Rust was to update the colors to the above.
Thanks!

PR #163, opened 2018-07-18
Branding colours updated

frano-m:
Primary, secondary, accent updated
#136


NoopDog:
@frano-m this is merged

PR #239, opened 2018-09-05
Content line widths

frano-m:
- reduced maximum line widths to improve readability of text (aligns with material design specs https://material.io/design/typography/understanding-typography.html#readability)


NoopDog:
this is merged

PR #270, opened 2018-10-12
Explore search links

frano-m:
- fix for data-portal-content ticket 149 (click through to from organ or graphs is broken by the change to the url structure)


NoopDog:
manually merged

PR #282, opened 2018-10-29
Modifications to search useability

frano-m:
- search disabled until data available
- selecting a suggestion takes user to explore without the need for pressing 'search' button


NoopDog:
merged

PR #459, opened 2019-07-22
Metadata Dictionary: Add copy to clipboard function to name and path.

### LINKED ISSUE #435: Metadata Dictionary: Add copy to clipboard function path on all entities

frano-m:
Resolves #435.


MillenniumFalconMechanic:
Thanks @frano-m.

PR #460, opened 2019-07-23
Add 'Draft Mode' for documents. Resolves #444.

### LINKED ISSUE #444: Add "Draft Mode" for documents
Add a field to the front matter

```
---
path: "/guides/userguides/why-i-left-the-woods"
date: "2019-05-14"
title: "Why I Left the Woods"
subTitle: ""
draft: true
---
```

Then for local, and dev environments show the "draft" = true items in the navigation if they are added to the sitemap.

For integration, staging, and prod, do not show the draft documents in the navigation even if they are listed in the site map.

This lets us avoid having branches per environment and the associated merging while also enabling us to create stub doc for people to work on.

So that we do not have to go back and add the `draft` to all the other documents, if it is not specified on a doc, we assume its "false".  So any existing docs will show in all environments. 

We will still need to use pull requests for modifying documents. The "Draft Mode" will smooth out the flow of having new docs sit on a branch until they are completed and reviewed.

frano-m:


NoopDog:
@frano-m @MillenniumFalconMechanic lets review this one together on a call.

MillenniumFalconMechanic:
Thanks @frano-m.

PR #461, opened 2019-07-23
Bug fix conflicting banner messages with mobile menu.

frano-m:
- fix for when banner is active but mobile menu open
- fix for header background on home page when scrolling


MillenniumFalconMechanic:
Good one @frano-m, thanks.

PR #1056, opened 2021-08-31
Lk contribution update

ekiernan:

### DIFF
--- a/content/contribute/data/contributing-to-hca.md
+++ b/content/contribute/data/contributing-to-hca.md
@@ -55,3 +55,34 @@ Learn more about what results will be available for different data types [here](
 > #### Start the Process
 > Start the HCA submission process by emailing\
 > [wrangler-team@data.humancellatlas.org](mailto:wrangler-team@data.humancellatlas.org).
+
+## Wrangler Office Hours
+
+If you have questions about data contribution, join the DCP Wranlger office hours on Zoom.
+
+**Date/time:** 
+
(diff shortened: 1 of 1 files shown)


content/contribute/data/contributing-to-hca.md
```diff
@@ -55,3 +55,36 @@ Learn more about what results will be available for different data types [here](
...
 > [wrangler-team@data.humancellatlas.org](mailto:wrangler-team@data.humancellatlas.org).
+
+## Wrangler Office Hours
+
+If you have questions about data contribution, join the DCP Wranlger office hours on Zoom.
+
+**Date/time:** 
+
+- Every 3rd Wednesday of the month at 4 pm **(PT)**, starting September 15, 2021
+- Every 2nd Friday of the month at 10 AM **(UTC +1)**, starting September 17, 2021
```
ESapenaVentura:
```suggestion
- Every 3rd Wednesday of the month at 11 pm **(UTC)**, starting September 15, 2021
- Every 2nd Friday of the month at 9 AM **(UTC)**, starting September 17, 2021
```

Since we are trying to reach worlwide, I think getting it in UTC time for a standard seems right, although I have to say I am not super convinced of how useful will it be for users (Will they be confused by UTC? is it ok to put the times in 2 different notations and for people to translate, or is it better to get it all in one notation and we asume they'll understand it?)

ESapenaVentura (commented):
Small comment on the times, other than that LGTM 👍

NoopDog (approved):

PR #1151, opened 2021-10-11
Add workflow to test the build

Arctenik:

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,23 @@
+name: Test build
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: main
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build.yml
```diff
@@ -0,0 +1,23 @@
...
+          repository: HumanCellAtlas/metadata-schema
+          path: hca-metadata-schema
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.16.0"
+      
+      - run: |
+          cd main
+          npm ci
+          npm install --global gatsby-cli
```
NoopDog: is the gatsby cli version specified in the package.json and can we use that instead of doing this global install with the latest version?

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,23 @@
...
+          source ./set-dev-env.sh && gatsby build --verbose
```
NoopDog: GitHub is warning there is no newline at the end of the file.

NoopDog (changes requested):

NoopDog (approved):

PR #1709, opened 2022-08-26
Switched banners from using cookies to local storage #1703

### LINKED ISSUE #1703: Convert privacyAccepted environmentAccepted cookes to local storage
#### Need
The security scan is asking for these cookies to be made inaccessible to javascript, but javascript needs these cookies to determine if we need to show the "Accept Privacy" banner or the "This is a test environment" banner.

#### Approach
We can address this by converting these cookies to local storage. They do not need to go to a server as these are not processed by any back end at the moment. 

On accept, instead of setting a cookie, set local storage. 
On load, instead of looking for a cookie, query the local storage to determine if the user has accepted the cookie.

Cookies affected:
privacyAccepted
environmentAccepted

mrtopsyt:
Switched the test and privacy banners to use local storage instead of cookies to determine whether they should show or not

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11044,11 +11044,6 @@
         }
       }
     },
-    "js-cookie": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/js-cookie/-/js-cookie-3.0.1.tgz",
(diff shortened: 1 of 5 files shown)


frano-m (changes requested):
Nice work @jpaten! Would you mind uninstalling `js-cookie` as we now no longer use it. Try `npm uninstall --save js-cookie`. Thank you so much!

frano-m (approved):
LGTM!

Issue #2233, opened 2023-10-27
[Atlas Pages] Scrape BICCN Publication RHS data


NoopDog:
You could download the HTML from the [original site](https://www.biccn.org/science/human-and-nhp-cell-atlas#primate-dev) and so some regex magik or filtering to pull out a data model that could be used to populate the RHS boxes for each DOI.

![image](https://github.com/DataBiosphere/data-portal/assets/3239697/6402d212-8fed-4780-86d6-7b09d7a8c03f)

Issue #2226, opened 2023-10-26
 [Atlas Pages} Read publications from cross ref and save title and authors


NoopDog:
We are doing a version of this lovely page here...
https://www.biccn.org/science/human-and-nhp-cell-atlas#primate-dev

biccn.orgbiccn.org
[Human and Non-Human Primate Cell Atlas - Brain Cell Data Center (BCDC)](https://www.biccn.org/science/human-and-nhp-cell-atlas#primate-dev)
Explore the human and non-human primate brain cell atlas produced by the BRAIN Initiative Cell Census Network (BICCN) with transcriptomics and other tools.
[7:54](https://clevercanary.slack.com/archives/C03KTB3DJ21/p1698288896545169)
The mocks are here
Dekstop: https://www.figma.com/file/i4uZsYkKAML9ivi2TIo6ug/HCA-Data-Explorer?type=design&node-id=5532%3A6559&mode=design&t=rDatfWbma0ADn48m-1
Mobile: https://www.figma.com/file/i4uZsYkKAML9ivi2TIo6ug/HCA-Data-Explorer?type=design&node-id=5480%3A6541&mode=design&t=rDatfWbma0ADn48m-1

[7:55](https://clevercanary.slack.com/archives/C03KTB3DJ21/p1698288938567689)
Fran did the implementation (without the categories so far) here:
https://data.humancellatlas.dev.clevercanary.com/bio-networks/nervous-system/publications
[7:55](https://clevercanary.slack.com/archives/C03KTB3DJ21/p1698288956867219)
Fran just did one publication with a hand copied version of one paper. (edited) 
[7:56](https://clevercanary.slack.com/archives/C03KTB3DJ21/p1698288984729069)
But there are 25 papers.
[7:58](https://clevercanary.slack.com/archives/C03KTB3DJ21/p1698289113271089)
So I was thinking
To get the title and authors, we could use your crossref code (we as in you).  An you would take a list of DOIs and expand them to a local database that  we check in like in our /files type work. This would just be a map that is keyed off of doi and has values of title and Authors (or whatever makes sense e.g first author or.. )

NoopDog:
Complete

PR #2234, opened 2023-10-30
feat: add biccn publications (#2226, #2233)

hunterckx:
Notes:
- In addition to the data being added, the MainColumn component has been updated to use it; categories are in the data but not implemented in the component (are there existing components that should be used for the category layout?)
- Apparently the titles returned from Crossref can have markup in them:
  `Human voltage-gated Na <sup>+</sup> and K <sup>+</sup> channel properties underlie sustained fast AP signaling`
  This is currently left as-is because I'm not sure what the best way to handle it would be
- I'm not sure if the HCA page was meant to include all the same information as the middle column of the [original BICCN page](https://www.biccn.org/science/human-and-nhp-cell-atlas), but I have made it mostly the same; differences are:
  - Author names are listed in `[given name] [surname]` format, as in the placeholder/mocks (and the list is ended with a period)
  - bioRxiv is listed like a journal rather than alongside DOI
- For the plural words for the n-tags in the right-hand column, I've just used plurals of the labels (e.g. portal => portals, tools => tools), except for the ones that are shown in the mock
- All of the publication information is in a single JSON file


frano-m (approved):
@hunterckx looks fantastic to me! Thank you.

Issue #2465, opened 2024-09-28
Resolve timeout when building network pages


hunterckx:
Issue appears on branch `fran/2359-retina`

NoopDog:
@frano-m can you base this in main so we can merge this in without merging in the atlases?

PR #2466, opened 2024-09-28
fix: request individual cellxgene collections to avoid timeout on datasets (#2465)

hunterckx:


frano-m (approved):
Brilliant @hunterckx thank you

PR #2471, opened 2024-10-07
chore: added id to 'Explore Data' button (#2469)

### LINKED ISSUE #2469: Track clicks on "Explore Data" CTA
#### Need
We need to add tracking of the primary CTA button on the homepage to monitor user interactions and optimize conversions. This tracking will allow us to analyze the conversion funnel more accurately, ensuring that the CTA is performing as expected.

#### Approach
* Add a `button-explore-data` ID to the "Explore Data" button.
* Use GTM's built-in click variables to enable click tracking.
* Create trigger that fires on "Click - All Elements" / "Some Clicks" to target only the "Explore Data" button.

#### Definition of Done
Clicks on "Explore Data" are accessible in the query explorer for both prod and staging:
- [x] data.humancellatlas.org - GA4
- [x] data.humancellatlas.dev.clevercanary.com - GA4

mrtopsyt:
### Changes
- Added an id reading `button-explore-data` to the "Explore Data" button on the index page

### DIFF
--- a/next/components/Home/components/SectionHero/sectionHero.tsx
+++ b/next/components/Home/components/SectionHero/sectionHero.tsx
@@ -34,7 +34,7 @@ export const SectionHero = (): JSX.Element => {
         </Head>
         <Subhead>Community generated, multi-omic, open data</Subhead>
         <CTAs>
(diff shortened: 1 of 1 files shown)


frano-m (approved):
Thank you @jpaten !

PR #2603, opened 2024-12-06
chore: update dev dependencies to latest

### LINKED ISSUE #2600: Update dev dependencies to latest
#### Need
We have quite a few dev dependencies that have dependabot alerts that they have later versions and many of these versions are major version bumps that might require changes in the various configuration files.

#### Approach

Bump the versions to the latest where possible and modify the configuration files accordingly to keep the functionality consistent.

- Work with the team to resolve any new eslint errors.
- Review any prettier formatting changes to see if we want to configure them away (if possible) or take them.
- Update the build script command and GitHub actions as required.
- Test that the app builds and functions okay after the build.

AnVIL portal recently had its dev dependencies updated so we can use that as a guide.
https://github.com/anvilproject/anvil-portal/pull/3428
https://github.com/anvilproject/anvil-portal/pull/3428/commits

mrtopsyt:
#### Changes
- Updated dev dependencies as specified by Dependabot (see comments in #2600)
- Made SonarJS linting changes
  - Set `sonarjs/todo-tag` and `sonarjs/redundant-type-aliases` as warns rather than errors
  - Added `eslint-disable` blocks to avoid errors due to `sonarjs` failing to detect `noopener noreferrer` when passed using a constant
- Made Prettier linting changes
  - Set prettier to use `es5` comma rules
  - Reran prettier to comply with the new version
- Updated hooks to avoid Husky deprecation warnings
- Moved `slugify`, `remark-gfm` and `next-mdx-remote` from `devDependencies` to `dependencies` since they were used in the app. Did not update these dependencies, as doing so led to typing errors


mrtopsyt:
Running list of changes
- sonar
  - noopener error
  - todo warn
  - redundant type alias warn
- prettier
  - es5 commas
  - commas in arrow brackets `<>` that run onto multiple pages

package.json
```diff
@@ -38,39 +38,39 @@
...
+    "@commitlint/cli": "^19.6.0",
+    "@commitlint/config-conventional": "^19.6.0",
+    "@digitak/esrun": "^3.2.26",
     "@mui/types": "^7.2.18",
-    "@next/eslint-plugin-next": "^14.1.0",
+    "@next/eslint-plugin-next": "^15.0.3",
     "@types/mdx": "^2.0.11",
-    "@types/node": "^20.10.0",
-    "@types/react": "^18.3.5",
+    "@types/node": "^22.10.1",
```
hunterckx: `@types/node` should be matched with the Node version, so I think this should stay at `^20.10.0` (and I believe this is a convention for other type packages, so we might want to keep some of the others the same as before as well)

hunterckx (changes requested):

mrtopsyt:
@hunterckx I updated the types dependencies to match the version of the corresponding prod dependency, or to the version they were at originally if there is there isn't a version of the type dependency that's the same as the prod dependency. Let me know if that works, I can also just change them to be the exact same as they were originally

hunterckx (approved):
Looks good!

PR #3109, opened 2026-07-01
fix: source Node version from .nvmrc in cc-data and cgl-data build scripts

### LINKED ISSUE #3108: Update Node version in the "n" step of the cc-data and cgl-data build scripts
## Problem

The two deploy/build scripts hardcode a Node version in the `n` step, and they are out of sync with each other and with the repo's pinned version.

- `cc-data.humancellatlas.dev.next.sh` line 11: `n 22.13.0`
- `cgl-data.humancellatlas.prod.next.sh` line 11: `n 22.12.0`

The repo pins Node `22.13.0` in both `.nvmrc` and `package.json` `engines`. `cgl-data` (prod) is building on `22.12.0`, which does not match.

## Proposed fix

Update the `n` step in both scripts to set the Node version properly so it matches the version pinned in `.nvmrc` / `package.json` engines (`22.13.0`), and stays consistent across both scripts.

Ideally the version should be sourced from `.nvmrc` rather than hardcoded, e.g.:

```sh
n "$(cat .nvmrc)"
```

so the scripts stay in sync automatically going forward.

NoopDog:
Closes #3108

## Problem

The `n` step in the two deploy/build scripts hardcoded a Node version and had drifted out of sync:

- `cc-data.humancellatlas.dev.next.sh` (dev) — `n 22.13.0`
- `cgl-data.humancellatlas.prod.next.sh` (prod) — `n 22.12.0`

The repo pins Node `22.13.0` in `.nvmrc` and `package.json` engines, so prod was building on a mismatched version.

## Change

Both scripts now source the version from `.nvmrc`:

```sh
n "$(cat .nvmrc)"
```

This makes both match the repo-pinned version and stay in sync automatically going forward. The scripts already run from the repo root, so the relative `.nvmrc` path resolves correctly.

### DIFF
--- a/cc-data.humancellatlas.dev.next.sh
+++ b/cc-data.humancellatlas.dev.next.sh
@@ -8,6 +8,7 @@ rm -rf ./out
 echo \"Deleting ./build/\"
 rm -rf ./build
 
+# Switch to the Node version pinned in .nvmrc (kept in sync with package.json engines)
 n "$(cat .nvmrc)"
 npm ci
 
--- a/cc-data.humancellatlas.dev.next.sh
+++ b/cc-data.humancellatlas.dev.next.sh
@@ -8,7 +8,7 @@ rm -rf ./out
 echo \"Deleting ./build/\"
 rm -rf ./build
 
-n 22.12.0
+n "$(cat .nvmrc)"
 npm ci
 
 mkdir -p build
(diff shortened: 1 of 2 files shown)


frano-m (approved):
Lovely @NoopDog

Issue #3199, opened 2026-09-08
feat: atlas source datasets table: add a Primary Data column linking to the HCA Data Explorer


NoopDog:
Part of #3193.

## Problem

Every source dataset derives from a source study whose primary data is often already in the HCA Data Explorer. The Source Studies tab exposed that as a bare external-link icon under an "HCA Data Explorer" heading; with that tab removed (#3194), the source datasets table offers no route from a dataset to its primary data.

Add a **Primary Data (fastqs)** column linking each row to the HCA Data Explorer project for its source study, presented as a branded portal link rather than a bare icon.

## Depends on

#3197 — it joins the source study onto each source dataset at build time and puts `hcaProjectId` on `TrackerSourceDataset`. Do not re-implement that join here.

## Approach — presentation

Render it the way the **Explore** column already renders CAP: the portal's mark followed by a short label, via the existing analysis-portal machinery, not as a bare `OpenInNewIcon`. Concretely: the HCA blue Voronoi-sphere mark with the label **Explorer**.

The column header names what is linked, and the cell names where it goes — so the header is `Primary Data (fastqs)` and the cell reads "Explorer".

That machinery is already in place and needs one new portal registered:

- `@types/network.ts` — add a member to the `ANALYSIS_PORTAL` enum for the HCA Data Explorer.
- `constants/analysisPortals.ts` — add an `HCA_DATA_EXPLORER` entry alongside `CAP`, `CZ_CELLXGENE`, and `UCSC_CELL_BROWSER`: `{ icon: "/hca-bio-networks/network/atlas/icons/<asset>", label: "Explorer", name: ANALYSIS_PORTAL.HCA_DATA_EXPLORER }`.
- `utils/network.ts` — add a builder beside `buildCAPAnalysisPortal`, returning `{ url, ...HCA_DATA_EXPLORER }`.
- Render with `C.AnalysisPortalCell({ analysisPortals })`, which maps each portal through `IconLink` (mark at `height={20}` plus the label as a link) — identical treatment to the Explore column.

**Asset needed.** There is no standalone blue-sphere icon in the repo. `public/hca-bio-networks/logos/logoHca.png` and `logoHumanCellAtlas.png` are full lockups (sphere + wordmark), not the mark on its own. Add the mark to `public/hca-bio-networks/network/atlas/icons/`, preferably as SVG to match `cap.svg` and `ucsc-cell.svg`, sized to read cleanly at 20px.

## Approach — the column

[...]

Issue #3194, opened 2026-09-08
feat: atlas pages: remove the Source Studies tab from tracker-sourced atlases


NoopDog:
Part of #3193.

## Problem

Tracker-sourced atlases show three tabs — Atlas Overview, Source Datasets, Source Studies — mirroring the tracker's admin UI rather than what a portal visitor needs. The Source Studies tab adds little: every source study already appears in the source datasets table as a DOI-linked citation (`SOURCE_STUDY` column), and its three remaining exclusive fields are moving onto the source datasets table under the sibling issues in #3193 (Journal and Reference Author as filters; `hcaProjectId` as the Primary Data column).

## Changes

- `components/HCABioNetworks/Network/Atlas/components/common/Tabs/tabs.tsx` — drop the Source Studies entry from `TRACKER_TABS`, leaving Atlas Overview + Source Datasets. **Leave `BASE_TABS` alone** — non-tracker atlases keep their "Source Studies" tab.
- `pages/hca-bio-networks/[network]/atlases/[atlas]/source-studies.tsx` — delete. It carries its own `getStaticPaths`/`getStaticProps` rather than going through `utils/atlasPages.ts`, so nothing else depends on them — but see "lift before deleting" below.
- `components/HCABioNetworks/Network/Atlas/components/SourceStudies/` — delete the whole tree (main column, tracker view, and the table's `columns.ts`, `hook.ts`, `accessor.ts`, `viewBuilder.tsx`, `types.ts`).
- `constants/routes.ts` — remove `NETWORK_ATLAS_SOURCE_STUDIES_PATTERN`.

## Lift before deleting

`source-studies.tsx` holds the only `getStaticPaths` in the repo that filters to *published tracker atlases* — iterating `NETWORKS`, skipping atlases with no `atlas.tracker`, and gating the rest on `isTrackerAtlasPublished(shortNameSlug, version)`. #3198 needs exactly that filter for the new `/source-datasets` route. Lift it into `utils/atlasPages.ts` as a shared helper as part of this deletion rather than letting it go and rewriting it there.

## Explicitly keep

- `content/HCABioNetworks/Network/Atlas/SourceStudies/description.mdx` and its `SourceStudiesDescription` export in `content/index.tsx` — still rendered by the **non-tracker** branch of `Datasets/components/MainColumn/mainColumn.tsx`. Deleting it breaks every Azul-sourced atlas page.
[...]

Issue #3193, opened 2026-09-08
[Epic] Atlas pages: consolidate the tracker-sourced atlas view onto one Source Datasets tab


NoopDog:
## Summary

Breast v1.0 (iHBCA v1) is the first atlas whose page content is sourced from the [HCA Atlas Tracker](https://tracker.data.humancellatlas.org) API rather than from `constants/networks.ts` and Azul (#3150). Liver v1.0 (#3178) and Gut v1.0 (#3177) followed, and atlases added from here on will be tracker-sourced.

That view was built by mirroring the shape of the tracker's own admin UI: a **Source Datasets** tab and a **Source Studies** tab. In the portal the source studies list earns little on its own — every source study is already represented in the source datasets table through its citation — so this epic collapses the two tabs into one, moves the fields that only existed on the source studies tab onto the source datasets table, and tightens how that table presents each row.

This is an **epic**. Each piece of work is tracked in a sub-issue linked below.

## Scope — tracker-sourced atlases only

Everything here applies **only** to atlases with `atlas.tracker` set in `constants/networks.ts` — today Breast v1.0, Liver v1.0, and Gut v1.0, plus every atlas added from here on.

Atlases still sourced from Azul are untouched, including their URLs: they keep `/datasets`, keep the "Source Studies" tab label, and keep the projects table. #3198 splits the shared route rather than renaming it, so nothing established moves.

## Sub-issues

| # | Change |
| --- | --- |
| #3194 | Remove the Source Studies tab from tracker-sourced atlases |
| #3200 | Derive the tracker `-r<revision>` file name from one shared helper |
| #3195 | Stack the revisioned file name over the source study in one pinned column |
| #3196 | External-link icon on the source study citation |
| #3197 | Filter by Journal and Reference Author |
| #3199 | Primary Data column linking to the HCA Data Explorer |
| #3198 | Serve tracker-sourced atlases at `/source-datasets` |

## Groundwork shared by two sub-issues

The tracker's source-dataset records (`GET /api/atlases/{id}/source-datasets`) carry no journal, no author, and no HCA project ID. Those live on the source-study record (`GET /api/atlases/{id}/source-studies`), joined by `sourceStudyId`. Confirmed against the live tracker for Breast v1.0.

Both the journal/author filters and the Primary Data column need that join, so it is introduced once — in #3197, at build time in `utils/trackerAtlasPages.ts`, which already fetches both lists in the same `Promise.all` — and consumed by #3199.

[...]

PR #3202, opened 2026-09-09
feat: atlas source datasets table: filter by journal and reference author (#3197)

### LINKED ISSUE #3197: feat: atlas source datasets table: filter by Journal and Reference Author
Part of #3193.

## Problem

**Journal** and **Reference Author** were filterable on the Source Studies tab — filter-only columns, hidden from the table via `columnVisibility: { journal: false, referenceAuthor: false }`. That tab is being removed (#3194) and the filters have not been carried across, so the source datasets table offers no way to narrow by journal or by author.

## Constraint

The tracker's source-dataset endpoint returns neither field. Verified against the live tracker for Breast v1.0 — a source-dataset record carries `publicationString` (`"Gray (2022) Developmental Cell"`), `sourceStudyId`, and `sourceStudyTitle`, while `journal` and `referenceAuthor` exist only on the source-study record.

Parsing `publicationString` is not an acceptable substitute: the tracker's own citation builder emits `Author (Year) Journal` only when both a publication date and a journal are present, falling back to `Author` or `Author - Unpublished` otherwise. Join on `sourceStudyId` instead.

## Approach

`getTrackerContentStaticProps` in `utils/trackerAtlasPages.ts` already fetches source datasets and source studies in the same `Promise.all`, and already enriches each source dataset with `datasetAsset`. Extend that step with the join:

```ts
const sourceStudyById = new Map(sourceStudies.map((s) => [s.id, s]));

const trackerSourceDatasets = sourceDatasets.map((sd) => {
  const sourceStudy = sourceStudyById.get(sd.sourceStudyId);
  return {
    ...sd,
    datasetAsset: buildTrackerSourceDatasetAsset(sd),
    hcaProjectId: sourceStudy?.hcaProjectId ?? null,
    journal: sourceStudy?.journal ?? null,
    referenceAuthor: sourceStudy?.referenceAuthor ?? null,
  };
});
```

Add the three fields to `TrackerSourceDataset` in `@types/network.ts`. `hcaProjectId` is unused by this issue — it is what the Primary Data issue consumes, and carrying it here keeps the join in one place rather than duplicating it. `?? null` rather than leaving fields undefined, because `getStaticProps` cannot serialize `undefined`.

Then add the filter-only columns, mirroring what the source studies table did, in `components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/`:

- `columns.ts` — `JOURNAL` (`accessorFn: (row) => row.journal ?? "Unpublished"`, `id: "journal"`, header `"Journal"`, `enableColumnFilter: true`, `filterFn: "arrIncludesSome"`) and `REFERENCE_AUTHOR` (`accessorKey: "referenceAuthor"`, `id: "referenceAuthor"`, header `"Reference Author"`, same filter config). Add both to `COLUMNS`.
- `utils.ts` — extend `getColumnVisibility` to return `journal: false` and `referenceAuthor: false` alongside the existing `integratedObject: false`.
- `meta.ts` — add `{ key: "journal", label: "Journal" }` and `{ key: "referenceAuthor", label: "Reference Author" }` to `META.categoryGroups[0].categoryConfigs`, so they appear in the filter sidebar.

The `"Unpublished"` fallback matches what the source studies table did. A source dataset whose `sourceStudyId` has no match in the source-studies response should not crash the build — decide whether it facets as `"Unpublished"`/unspecified or is left out, and make it explicit.

## Acceptance criteria

- [ ] **Journal** and **Reference Author** filters appear in the source datasets filter sidebar for tracker-sourced atlases, alongside the existing Assay, Tissue, Disease, Source Study, and Integrated Object filters.
- [ ] Selecting a journal narrows the table to the source datasets whose source study was published in that journal; selecting an author does the same. Facet counts match the filtered row counts.
- [ ] Selections render as removable tags via `ColumnFilterTags`, and clearing them restores the full table.
- [ ] **No Journal or Reference Author column is displayed** in the table itself — filter-only, as on the old Source Studies tab.
- [ ] A source study with no journal facets as **Unpublished**.
- [ ] Verified on Breast v1.0: 7 source datasets across 7 source studies, with journals Nature, Developmental Cell, Cell Systems, Nature Genetics (×2), Nature Communications, and The EMBO Journal; authors Kumar, Gray, Murrow, Nee, Reed, Twigger, Pal.
- [ ] `TrackerSourceDataset` carries `journal`, `referenceAuthor`, and `hcaProjectId`, and `npx tsc --noEmit` passes.
- [ ] `npm run build-prod:data-portal` succeeds — the join runs at build time and must not throw on a missing source study.

## Notes

Blocks the Primary Data column issue in this epic, which consumes `hcaProjectId` from this join. Best landed after #3194.

frano-m:
## Summary

Closes #3197. Part of #3193.

**Journal** and **Reference Author** were filterable on the removed Source Studies tab and had not been carried across. The tracker's source-dataset endpoint returns neither field — verified against the live tracker, a source-dataset record carries `publicationString`, `sourceStudyId` and `sourceStudyTitle`, while `journal` and `referenceAuthor` exist only on the source-study record. So this joins the source study onto each source dataset at build time and adds both as filter-only columns.

## Changes

**The join** — `utils/trackerAtlasPages.ts`. `getTrackerContentStaticProps` already fetched both collections in one `Promise.all` and already enriched each dataset with `datasetAsset`; that step now also joins by `sourceStudyId`. `hcaProjectId` rides along for #3199 so the join lives in one place. `?? null` throughout, since `getStaticProps` cannot serialize `undefined`.

**The filters** — `Datasets/.../tracker/components/table/`:
- `accessor.ts` — `buildJournal` / `buildReferenceAuthor`, matching the existing `buildIntegratedObjects` pattern.
- `columns.ts` — `JOURNAL` and `REFERENCE_AUTHOR`, both `enableColumnFilter` with `filterFn: "arrIncludesSome"`.
- `utils.ts` — both hidden via `getColumnVisibility`, alongside `integratedObject`.
- `meta.ts` — both registered in `META.categoryGroups[0].categoryConfigs` for the sidebar.

**Types** — `@types/network.ts` adds `journal`, `referenceAuthor` and `hcaProjectId` to `TrackerSourceDataset`.

## Also in this PR

Two adjacent changes, each its own commit:

- **`cc36ca41` — drops the readerless `trackerSourceStudies` static prop and `AtlasContext` field.** #3194 deliberately kept these and designated #3197 as what replaces them; the per-dataset join now does. Nothing read them. **Saves 36,530 bytes of `__NEXT_DATA__` per tracker atlas page** (`datasets.html` 152,535 → 116,005). The fetch itself stays — the join needs it.
- **`bcac644f` — collapses the column filters into the drawer below 1024px, up from 820px.** The chip row measures ~820px in the browser, which was *also* the old breakpoint, so the chips overflowed the viewport at any width just above it once these two filters were added. The value is now `FILTER_DRAWER_BREAKPOINT` in `components/common/Filters/constants.ts`, shared by `ColumnFilters` and `ColumnFilterTags` so the chips and their selected-value tags always switch together.

## Verification

Against live tracker data for Breast v1.0, read back out of the built `__NEXT_DATA__`:

- 7 source datasets ↔ 7 source studies, **zero unmatched `sourceStudyId`**.
- Journals: Nature, Developmental Cell, Cell Systems, Nature Genetics (×2), Nature Communications, The EMBO Journal. Authors: Kumar, Gray, Murrow, Nee, Reed, Twigger, Pal. Both match the issue exactly.
- `hcaProjectId` present on all 7, `null` for Reed — the case #3199's empty-cell criterion covers.
- Chip row width measured in-browser at 815–820px (Assay 85, Tissue 88, Disease 97, Source Study 134, Integrated Object 160, Journal 94, Reference Author 162).
- `npm run lint` (0 errors), `npm run check-format`, `npx tsc --noEmit`, `npm run build-prod:data-portal` (64/64 pages) all pass.

## Notes for review

- **A join miss cannot hide a dataset.** An unmatched `sourceStudyId` leaves all three fields `null` rather than dropping the row: journal facets as `"Unpublished"`, author as `LABEL.UNSPECIFIED`. The labels differ deliberately — `journal` is legitimately null for an unpublished study, whereas the tracker always sets `referenceAuthor`, so a null there means only a join miss.
- `||` rather than `??` in both accessors, so an empty-string journal from Crossref cannot render a blank, unlabelled filter option.
- `hcaProjectId` has no consumer until #3199, as the issue specifies.
- **Pre-existing wart worth its own ticket, not addressed here:** `TrackerSourceStudy` models `journal` and `doi` as independently nullable, so a study that *is* published but whose Crossref record lacks a journal will facet as "Unpublished" while its Source Study cell renders a live DOI link. This PR preserves the old tab's behaviour, which the acceptance criteria require.
- The "fits at 1024" conclusion is arithmetic (820 needed vs ~980 available after gutters) — the 820px content width was measured directly; Fran verified filters fit.

### DIFF
--- a/@types/network.ts
+++ b/@types/network.ts
@@ -146,7 +146,6 @@ export interface AtlasContext extends NetworkContext {
   atlas: Atlas;
   projectsResponses: ProjectsResponse[];
   trackerSourceDatasets?: TrackerSourceDataset[];
-  trackerSourceStudies?: TrackerSourceStudy[];
 }
 
 export interface NetworkContext {
@@ -229,8 +228,11 @@ export interface TrackerSourceDataset {
   fileId: string;
   fileName: string;
   geneCount: number;
+  hcaProjectId: string | null; // Joined from the source study; null when the study has no HCA project or the join misses.
   id: string;
+  journal: string | null; // Joined from the source study; null when unpublished or the join misses.
   publicationString: string | null;
+  referenceAuthor: string | null; // Joined from the source study; null only when the join misses.
   revision: number;
   sizeBytes: number;
   sourceStudyId: string;
--- a/@types/network.ts
+++ b/@types/network.ts
@@ -241,6 +241,17 @@ export interface TrackerSourceDataset {
   title: string;
 }
 
+/**
+ * A source dataset as the tracker API actually returns it, before
+ * `getTrackerContentStaticProps` adds `datasetAsset` and joins the
+ * source-study fields on. Keeps the fetch signature honest so the joined
+ * fields cannot be read off an unenriched dataset.
+ */
+export type TrackerSourceDatasetResponse = Omit<
+  TrackerSourceDataset,
+  "datasetAsset" | "hcaProjectId" | "journal" | "referenceAuthor"
+>;
+
 export interface TrackerSourceStudy {
   cellxgeneCollectionId: string | null;
   contactEmail: string | null;
--- a/components/common/Filters/components/ColumnFilters/columnFilters.tsx
+++ b/components/common/Filters/components/ColumnFilters/columnFilters.tsx
@@ -7,6 +7,7 @@ import { BUTTON_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/b
 import { Theme, useMediaQuery } from "@mui/material";
 import { RowData } from "@tanstack/react-table";
 import { ComponentProps, JSX } from "react";
+import { FILTER_DRAWER_BREAKPOINT } from "../../constants";
 import { StyledButton, StyledButtonGroup } from "./columnFilters.styles";
 import { Props } from "./types";
 import { buildColumnFilters } from "./utils";
@@ -15,7 +16,7 @@ export const ColumnFilters = <T extends RowData>({
   table,
 }: Props<T>): JSX.Element | null => {
   const isDrawer = useMediaQuery(
-    (theme: Theme) => theme.breakpoints.down(820),
+    (theme: Theme) => theme.breakpoints.down(FILTER_DRAWER_BREAKPOINT),
     { noSsr: true }
   );
 
--- a/components/common/Filters/constants.ts
+++ b/components/common/Filters/constants.ts
@@ -1,0 +1,4 @@
+// Viewport width below which column filters collapse from the inline chip row
+// into the filter drawer. Shared by `ColumnFilters` and `ColumnFilterTags` so
+// the chips and their selected-value tags always switch together.
+export const FILTER_DRAWER_BREAKPOINT = 1024;
--- a/apis/tracker/api.ts
+++ b/apis/tracker/api.ts
@@ -1,6 +1,6 @@
 import type {
   TrackerComponentAtlas,
-  TrackerSourceDataset,
+  TrackerSourceDatasetResponse,
   TrackerSourceStudy,
 } from "../../@types/network";
 import type { PublishedAtlas } from "./types";
@@ -58,7 +58,7 @@ export function fetchTrackerComponentAtlases(
  */
 export function fetchTrackerSourceDatasets(
   atlasId: string
-): Promise<TrackerSourceDataset[]> {
+): Promise<TrackerSourceDatasetResponse[]> {
   return fetchTrackerApi(
     `/api/atlases/${atlasId}/source-datasets`,
     "source datasets"
--- a/components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/accessor.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/accessor.ts
@@ -1,4 +1,9 @@
+import { LABEL } from "@databiosphere/findable-ui/lib/apis/azul/common/entities";
 import type { TrackerSourceDataset } from "../../../../../../../../../../../@types/network";
+
+// Facet value for a source study with no journal, matching the label the
+// removed Source Studies table used.
+const JOURNAL_UNPUBLISHED = "Unpublished";
 
 /**
  * Returns the names of the integrated objects (sourced from the legacy
@@ -8,4 +13,25 @@ import type { TrackerSourceDataset } from "../../../../../../../../../../../@typ
  */
 export function buildIntegratedObjects(row: TrackerSourceDataset): string[] {
   return row.componentAtlases?.map(({ name }) => name) ?? [];
+}
+
+/**
+ * Returns the journal of a source dataset's source study, falling back to
+ * "Unpublished" when the study has no journal (or the build-time join missed).
+ * @param row - Tracker source dataset.
+ * @returns journal name, or "Unpublished".
+ */
+export function buildJournal(row: TrackerSourceDataset): string {
+  return row.journal || JOURNAL_UNPUBLISHED;
+}
+
+/**
+ * Returns the reference author of a source dataset's source study. The tracker
+ * always sets this on a source study, so the fallback covers only a missed
+ * build-time join — hence "Unspecified" rather than "Unpublished".
+ * @param row - Tracker source dataset.
+ * @returns reference author, or "Unspecified".
+ */
+export function buildReferenceAuthor(row: TrackerSourceDataset): string {
+  return row.referenceAuthor || LABEL.UNSPECIFIED;
 }
--- a/components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/columns.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/columns.ts
@@ -10,7 +10,11 @@ import {
   buildPinnedNTagProps,
   renderPinnedNTagCell,
 } from "../../../../../../../../../../common/Table/components/Cell/components/PinnedNTagCell/utils";
-import { buildIntegratedObjects } from "./accessor";
+import {
+  buildIntegratedObjects,
+  buildJournal,
+  buildReferenceAuthor,
+} from "./accessor";
 import {
   renderCellCount,
   renderDownload,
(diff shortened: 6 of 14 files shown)


@types/network.ts
```diff
@@ -229,8 +228,11 @@ export interface TrackerSourceDataset {
   fileId: string;
   fileName: string;
   geneCount: number;
+  hcaProjectId: string | null; // Joined from the source study; null when the study has no HCA project or the join misses.
   id: string;
+  journal: string | null; // Joined from the source study; null when unpublished or the join misses.
   publicationString: string | null;
+  referenceAuthor: string | null; // Joined from the source study; null only when the join misses.
```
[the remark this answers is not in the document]
frano-m: Good catch — fixed in d0326c00.

You're right, and it's confirmed against the live endpoint: a source-dataset record from `/api/atlases/{id}/source-datasets` carries `publicationString`, `sourceStudyId` and `sourceStudyTitle`, but none of `journal`, `referenceAuthor` or `hcaProjectId`. So the fetch signature was asserting three fields the payload never returns.

Rather than making the three fields optional — which would push `undefined` handling into every consumer for values the join always sets — the API boundary is now typed separately:

```ts
export type TrackerSourceDatasetResponse = Omit<
  TrackerSourceDataset,
  "datasetAsset" | "hcaProjectId" | "journal" | "referenceAuthor"
>;
```

`fetchTrackerSourceDatasets` returns `TrackerSourceDatasetResponse[]` and `buildTrackerSourceDatasetAsset` takes the same type (a widening, so existing callers passing a full dataset still work). `datasetAsset` is in the `Omit` too, since it's added by the same enrichment step — that optional field was the existing precedent I should have followed. Everything downstream (the table, `AtlasContext`, the accessors) keeps consuming the enriched `TrackerSourceDataset`, so the UI retains its non-optional guarantee.

Verified the guard actually bites rather than just compiling: reading a joined field off an unenriched dataset is now a compile error —

```
error TS2339: Property 'journal' does not exist on type 'TrackerSourceDatasetResponse'.
```

`tsc --noEmit`, lint, prettier and `build-prod` all pass, and the generated output is unchanged (7 datasets, 6 journals, 7 authors, all three fields present).

components/common/Filters/constants.ts
```diff
@@ -0,0 +1,4 @@
+// Viewport width below which column filters collapse from the inline chip row
+// into the filter drawer. Shared by `ColumnFilters` and `ColumnFilterTags` so
+// the chips and their selected-value tags always switch together.
+export const FILTER_DRAWER_BREAKPOINT = 1024;
```
NoopDog, line 4: 1024 is already the design system's `sm` (`site-config/common/constants.ts:8`), so `theme.breakpoints.down("sm")` is identical today and can't drift. As a literal, the filter drawer would silently stop switching with the rest of the layout if `sm` is ever retuned.

components/common/Filters/components/ColumnFilters/columnFilters.tsx
```diff
@@ -15,7 +16,7 @@ export const ColumnFilters = <T extends RowData>({
   table,
 }: Props<T>): JSX.Element | null => {
   const isDrawer = useMediaQuery(
-    (theme: Theme) => theme.breakpoints.down(820),
+    (theme: Theme) => theme.breakpoints.down(FILTER_DRAWER_BREAKPOINT),
```
NoopDog, line 19: Worth a re-measure: the ~820px chip row was measured with no filters selected. `FilterCountChip` returns `null` at count 0, so each chip gains the badge (`min-width: 16px` + `padding: 0 4px`) plus the 8px gap once its filter is active — ~24px per selected filter. Seven active is ~820 + 168 ≈ 988px against the ~980px of content width at a 1024px viewport, and `StyledButtonGroup` is a MUI `ButtonGroup` with no wrap — the same overflow this commit fixes. It may just clear, but the headroom is thinner than the measurement suggests.

NoopDog (approved):
Approving — the join is sound and the verification in the description is unusually thorough.

`arrIncludesSome` on a scalar accessor was what I most expected to bite: TanStack's built-in substring-matches, which would make selecting "Nature" also match "Nature Genetics" and "Nature Communications" in your own sample data. But `COLUMN_FILTERS_OPTIONS.filterFns` supplies findable-ui's version, which uses `===` for non-arrays, and `table.options.filterFns` takes precedence over the built-ins. Exact match confirmed.

Also checked: hidden columns still surface in both the chip row and the drawer (`ColumnFiltersAdapter` filters on `getCanFilter()`, not visibility) and stay out of the grid; the `meta.categoryConfigs` keys match the new column ids; deselecting the last value yields `undefined` rather than `[]`, so there's no empty-array wipeout; and `trackerSourceStudies` has zero readers repo-wide, so `cc36ca41` is safe. `d0326c00` already covers the one type concern I had.

Two non-blocking comments inline.

components/common/Filters/constants.ts
frano-m: Keeping the literal — `down("sm")` isn't equivalent here, and would be a regression.

The `sm: 1024` in `site-config/common/constants.ts` is applied by the nested `ThemeProvider` in `pages/_app.tsx`, and that provider wraps **only `<Header {...header} />`**. `<Main>` and the page component sit outside it, so content renders against the outer `appTheme`, where `sm` comes from findable-ui's `theme/common/breakpoints.js`:

```js
values: { lg: 1440, md: 1280, sm: 768, xs: 0, ... }
```

`md` and `lg` are identical in both, so that override exists purely to shift `sm` from 768 → 1024 for the header. `ColumnFilters` is in the content tree, so `theme.breakpoints.down("sm")` there resolves to **768**, not 1024 — which would leave the chip row (~950px with count badges) overflowing across the whole 768–1024 band, the exact bug this commit fixes.

On the drift argument, I'd actually invert it: staying decoupled from `sm` is the point. This threshold is a function of the chip row's own measured width, not of where the header chooses to collapse. If `sm` were ever retuned *downward*, binding to it would silently break the fit — so tracking `sm` introduces the drift risk rather than removing it. There's also no named token equal to 1024 in the content theme (0 / 768 / 1280 / 1440), so a named app-level constant is the honest way to express it.

When we add facets and the row outgrows this width, we'll adjust `FILTER_DRAWER_BREAKPOINT` deliberately, as a measured change.

components/common/Filters/components/ColumnFilters/columnFilters.tsx
frano-m: Re-measured at 1024 with every filter active — it fits, no change needed.

You were right that the badges materially widen the row; my ~820px figure was indeed taken with nothing selected. Measured in devtools at iPad Pro (1024 × 1366 CSS px) with **all seven filters active and badged** (Assay 2, Tissue 1, Disease 1, Source Study 2, Integrated Object 1, Journal 2, Reference Author 2):

- Chip row's right edge lands at **~950px** against the ruler — inside the viewport, no clipping and no horizontal scroll.
- So the badges add ~130px rather than ~168px, leaving roughly 70px of headroom instead of the ~-8px the estimate implied.

Confirmed the dropdowns still open and filter correctly in that state, and the selected-value tags wrap onto a second line below — that row is a `Grid`, not the `ButtonGroup`, so it wraps rather than overflowing.

Agreed the headroom is thinner than the original measurement suggested, and that this is the worst case for *this* atlas. Our position is that when we add facets we'll adjust `FILTER_DRAWER_BREAKPOINT` to match, re-measuring at that point rather than trying to pre-empt it.

frano-m: image

Issue #3203, opened 2026-09-09
fix: a failed published-atlases fetch can half-generate a tracker atlas and still exit 0


frano-m:
## Problem

`getPublishedAtlases` in `apis/tracker/api.ts` memoises the published-atlases list for the build, and on failure degrades to "nothing published" so a tracker outage cannot break non-tracker pages. But its `.catch()` does two things at once:

```js
.catch((err) => {
  console.warn(`Failed to fetch published atlases from tracker; treating as empty. ${err}`);
  publishedAtlasesPromise = null;   // <- clears the cache
  return [];
});
```

Returning `[]` **and** nulling the cached promise makes the publication gate non-deterministic *within a single build*: the first caller gets `[]`, and because the cache was cleared the next caller issues a fresh fetch that may succeed and return the real list. Two callers of the same gate can get different answers seconds apart.

Next calls `getStaticPaths` once per page module, and two of them consult the gate:

| Page | Paths function | Consults gate |
| --- | --- | --- |
| `index.tsx` | `getStaticPaths` | yes |
| `datasets.tsx` | `getNonTrackerStaticPaths` | no |
| `source-datasets.tsx` | `getTrackerStaticPaths` | yes |

## Failure modes

- **Gate fails for `source-datasets`, succeeds for `index`** — no `/source-datasets` page is written, but the atlas overview *is*. `Tabs` selects `TRACKER_TABS` from `atlas.tracker` being truthy and never consults publication state, so the overview ships a "Source Datasets" tab pointing at a page that does not exist. 404 on click.
- **The reverse** — `/source-datasets` is written with an "Atlas Overview" tab pointing at a missing page.

The damaging property is the silence: the build exits 0 with only a `console.warn` in the log, having half-generated an atlas.

## Not a regression

This is pre-existing, not introduced by #3198. Before that split, `index.tsx` and `datasets.tsx` both used the gated `getStaticPaths`, so there were already two independent gated calls that could diverge the same way. #3198 keeps the count at two — it only moves which page holds the second one.

## Suggested fix

Stop nulling the promise in the `.catch()`, so the empty result is cached for the build and every caller agrees. Consistently degraded is much safer here than inconsistent: an atlas is then either wholly present or wholly absent, never half.

Failing the build outright would be louder, but sacrifices the deliberate protection that keeps non-tracker pages building during a tracker outage — worth discussing, but the consistency fix addresses the broken-tab outcome on its own.

[...]

Issue #3200, opened 2026-09-08
refactor: derive the tracker -r<revision> file name from one shared helper


NoopDog:
Part of #3193.

## Problem

Tracker files are published under a revisioned name — `<stem>-r<revision><ext>`, e.g. `ihbca-v1-gray-2022.h5ad` at revision 1 becomes `ihbca-v1-gray-2022-r1.h5ad`. The revision comes from the tracker (`revision` on both source-dataset and component-atlas records).

That one convention is currently derived **three different ways**:

1. `utils/trackerNetwork.ts:163-169` — `buildVersionedFileName(baseFileName, revision)`, the canonical implementation. It is **not exported**, so its only caller is `buildTrackerDownloadUrl` building the S3 URL.
2. `components/HCABioNetworks/.../Datasets/.../tracker/components/table/viewBuilder.ts:38-41` — re-implements the suffix inline as `` `${stem}-r${revision}` `` for the source datasets download cell, splitting `baseFileName` itself and passing the extension separately as `format`.
3. `viewModelBuilders/viewModelBuilders.ts:100-106` — for integrated objects, takes a third route entirely: pops the last path segment off the already-built `asset.downloadURL` and splits that, reverse-engineering the name out of a URL rather than deriving it from `revision`.

They agree today. A change to the convention has to be made in three places, in two different styles, or the name shown in the table drifts from the file actually downloaded — and #3195 is about to add a fourth consumer, the File Name column.

This was called out as motivation in #3189 and listed as explicitly out of scope in #3190; this is the issue that fixes it.

## Approach

The versioned name is already computed once, at build time, inside `buildTrackerDatasetAsset` when it builds the download URL. Carry it on the asset instead of recomputing or re-parsing it downstream:

- `@types/network.ts` — add `fileName` to `DatasetAsset` (the versioned file name, with extension). Optional, since CXG assets are built elsewhere and rendered by `CXGDownloadCell`, which does not use it.
- `utils/trackerNetwork.ts` — export `buildVersionedFileName`, and set `fileName` on the asset returned by `buildTrackerDatasetAsset` using the value it already derives for the URL.
- `.../table/viewBuilder.ts` `renderDownload` — read the name from `datasetAsset` rather than rebuilding it. `TrackerDownloadCell` takes `fileName` and `format` separately, so keep using `splitFileName` to split the asset's name for display; just stop re-deriving the `-r<revision>` part.
[...]

Issue #3196, opened 2026-09-08
feat: atlas source datasets table: add an external-link icon to the source study citation


NoopDog:
Part of #3193.

## Problem

The source study citation links out to `https://doi.org/{doi}` in a new tab, but renders as plain link text with no affordance that it leaves the site. The (now removed) Source Studies tab already used `OpenInNewIcon` for exactly this, so the pattern and the icon both exist.

## Depends on

#3195 — that issue folds the source study citation into the stacked File Name cell as its secondary line. This issue adds the external-link affordance to that line, so it should land after.

## Approach

The citation link is built by `renderSourceStudy` in `components/HCABioNetworks/Network/Atlas/components/Datasets/components/MainColumn/tracker/components/table/viewBuilder.ts`, which passes the citation string as `LinkProps.label`. `label` is typed `ReactNode`, so it can carry the citation text followed by the icon.

That file is `.ts` and holds no JSX today. Rename it to `viewBuilder.tsx` and compose the label as a fragment — the source studies equivalent was already `.tsx`. The import in `columns.ts` is extensionless, so no import changes are needed. `OpenInNewIcon` is re-exported from `components/index.tsx` and reachable as `C.OpenInNewIcon`. (If #3195 moves this rendering into a dedicated cell component, apply the change there instead.)

While in here: the current implementation passes `url: ""` when the source dataset has no `doi`, producing a link that goes nowhere. Decide the no-DOI case explicitly — the recommendation is to render the citation as plain text with no link and no icon, since an external-link icon on a link to nowhere is worse than the status quo.

## Acceptance criteria

- [ ] Every source study citation that links to a DOI shows an external-link icon immediately after the citation text.
- [ ] The text and the icon are a single link target — clicking either opens the DOI.
- [ ] The icon is `OpenInNewIcon` from `components/index.tsx`, sized and aligned to the secondary (`body-small-400`) line it sits on, without changing the row height.
- [ ] The link keeps `target="_blank"` and `rel="noopener noreferrer"`.
- [ ] A row with a citation but no DOI renders as decided above — no link to nowhere, no icon.
- [ ] A row with no `publicationString` renders no second line at all, as #3195 specifies.
- [ ] Verified on Breast v1.0, where all seven source datasets carry a DOI.
- [ ] `npm run lint`, `npm run check-format`, and `npx tsc --noEmit` pass after any `.ts` → `.tsx` rename.

## Note

[...]

PR #3204, opened 2026-09-09
feat: atlas pages: serve tracker-sourced atlases at /source-datasets (#3198)

### LINKED ISSUE #3198: feat: atlas pages: serve tracker-sourced atlases at /source-datasets, leaving /datasets for the rest
Part of #3193.

## Problem

For tracker-sourced atlases the tab is labelled **Source Datasets** but is served from `/hca-bio-networks/{network}/atlases/{atlas}/datasets`. Rename the slug to `source-datasets` so the URL matches the tab, and so it reads as the counterpart to the source-studies concept it absorbs (#3194).

## Scope — tracker-sourced atlases only

**Non-tracker atlases keep `/datasets` exactly as it is.** Only Breast v1.0 and the tracker-sourced atlases that follow move to `/source-datasets`.

This matters because `pages/hca-bio-networks/[network]/atlases/[atlas]/datasets.tsx` currently serves both atlas types off one dynamic route: `getStaticPaths` in `utils/atlasPages.ts` emits every atlas, and `getContentStaticProps` branches on `atlas.tracker`, with `MainColumn` branching again to render either the tracker source datasets view or the Azul projects table. Scoping the rename to tracker atlases therefore means splitting that route in two rather than renaming it in place.

## No redirect needed

The tracker `/datasets` URLs are new — they have only existed since Breast v1.0 shipped (#3150), are reached through in-app tab navigation, and are not linked from outside the site. Nothing established moves, because the established `/datasets` URLs belong to non-tracker atlases and those are untouched.

So the static-export redirect problem does not arise here: there is no CloudFront or S3 work in this issue, and no stub page.

## Approach — split the route

- **New** `pages/hca-bio-networks/[network]/atlases/[atlas]/source-datasets.tsx` — tracker atlases only. Its `getStaticPaths` filters to atlases with `atlas.tracker` that pass `isTrackerAtlasPublished`, which is exactly the filter `source-studies.tsx` already implements today (worth lifting into `utils/atlasPages.ts` as a shared helper as that page is deleted by #3194). Static props come from `getTrackerContentStaticProps`. Renders the tracker `SourceDatasets` view.
- **Keep** `datasets.tsx` — non-tracker atlases only. Its `getStaticPaths` filters to atlases *without* `atlas.tracker`. Renders the Azul projects table.
- `constants/routes.ts` — keep `NETWORK_ATLAS_DATASETS_PATTERN` for the non-tracker route; add `NETWORK_ATLAS_SOURCE_DATASETS_PATTERN` = `NETWORKS_ATLAS_PATTERN + "/source-datasets"`.
- `components/.../common/Tabs/tabs.tsx` — `TRACKER_TABS` points its Source Datasets entry at the new constant; `BASE_TABS` is unchanged.

## Simplifications this unlocks

Each page now has exactly one kind of content, so two pieces of branching can go:

- `Datasets/components/MainColumn/mainColumn.tsx` — the `isTracker` branch. Give the tracker page its own main column rendering `SourceDatasets`, and leave the existing one rendering only the projects table with its MDX description.
- `getContentStaticProps` in `utils/atlasPages.ts` — the `nonTrackerTabName` fudge that rewrites the tab name "Source Datasets" back to "Source Studies" for non-tracker atlases. Each page can pass its own tab name directly.

Both are optional cleanups, but they are the reason this split is worth doing rather than a cost of it. If they are skipped, say so in the PR.

## Acceptance criteria

- [ ] Tracker-sourced atlases serve their source datasets tab at `/hca-bio-networks/{network}/atlases/{atlas}/source-datasets`.
- [ ] Non-tracker atlases are completely unchanged: still `/datasets`, still labelled "Source Studies", still rendering the Azul projects table.
- [ ] The static export emits `source-datasets/index.html` only for published tracker atlases, and `datasets/index.html` only for non-tracker atlases — no atlas emits both, and no atlas emits neither.
- [ ] Tab navigation, direct URL entry, and browser back/forward work on both routes, with the correct tab shown as selected (`Tabs` matches on `router.pathname`, so the route constants and the page filenames must agree).
- [ ] The generated sitemap lists each atlas's tab once, under the right slug.
- [ ] An unpublished tracker atlas generates neither route, as today.
- [ ] `npm run lint`, `npm run check-format`, `npx tsc --noEmit`, and `npm run build-prod:data-portal` all pass.

## Note

Lands after #3194, which removes the Source Studies tab and its page — both issues touch `tabs.tsx`, and #3194 deletes the `getStaticPaths` tracker filter that this issue wants to reuse.

frano-m:
## Summary

Closes #3198. Part of #3193.

Tracker-sourced atlases now serve their source datasets tab at `/hca-bio-networks/{network}/atlases/{atlas}/source-datasets`, so the URL matches the tab label. **Non-tracker atlases are completely untouched** — still `/datasets`, still labelled "Source Studies", still the Azul projects table. The shared route is split rather than renamed, so nothing established moves.

## ⚠️ Note for the epic: the tracker table components have moved

This is the one thing to be aware of before picking up the rest of #3193.

```
BEFORE  Datasets/components/MainColumn/tracker/components/table/
AFTER   SourceDatasets/components/MainColumn/components/table/
```

The tracker view and its table now live under a new `SourceDatasets/` sibling, alongside `Overview/` and `Datasets/`, so each route owns exactly one component tree. The redundant `tracker/` path level is gone, since the whole tree is now tracker-only.

**#3193's Sequencing section says steps 3–5 "touch only the table components under `Datasets/components/MainColumn/tracker/components/table/` — none of which #3198 moves".** That is no longer true. **#3200, #3195, #3196 and #3199 all cite the old path and will need the new one**, and they should now land *after* this PR rather than in parallel with it. I've commented on the epic and each of those four issues.

Every move is recorded by git as a rename (`R`), so history and blame follow the files.

## Changes

- **New** `pages/.../[atlas]/source-datasets.tsx` — tracker atlases only, via `getTrackerStaticPaths` (the helper #3194 lifted for exactly this).
- **Kept** `pages/.../[atlas]/datasets.tsx` — non-tracker only, via a new `getNonTrackerStaticPaths`.
- `constants/routes.ts` — adds `NETWORK_ATLAS_SOURCE_DATASETS_PATTERN`; `NETWORK_ATLAS_DATASETS_PATTERN` stays for the non-tracker route.
- `tabs.tsx` — `TRACKER_TABS` points at the new constant. `BASE_TABS` unchanged.

**Both optional cleanups from the issue are included:**

- `Datasets/.../mainColumn.tsx` — the `isTracker` branch is gone; it renders only the projects table and its MDX description.
- `getContentStaticProps` — the `nonTrackerTabName` fudge is gone. Each page passes its own tab name, which the page titles confirm.

`getStaticPaths`, `getNonTrackerStaticPaths` and `getTrackerStaticPaths` were three copies of one nested loop differing only in predicate, so they now share a `buildStaticPaths(predicate)` helper — the same drift risk raised on #3197, rather than a third copy of it. Also deleted a dead duplicate `Props` interface carried through the move (no importers).

## Verification

All against the real static export, not just the source:

- **Route split holds exactly**: `breast-v1` emits `source-datasets.html` only; the six non-tracker atlases emit `datasets.html` only. Programmatically checked that **no atlas emits both and none emits neither**.
- **Tab selection**: `aria-selected="true"` lands on *Source Datasets* on the new route and on *Source Studies* on `/datasets`, confirming the route constants and page filenames agree.
- **Titles** prove the fudge removal: `… - Source Datasets` vs `… - Source Studies`.
- **Sitemap**: 6 × `/datasets` + 1 × `/source-datasets`, each atlas once, no duplicates.
- **Non-tracker unchanged**: 34 projects on lung, MDX description and its HCA Data Explorer link intact.
- **Unpublished tracker atlases, with real data**: the tracker publishes only Breast v1.0, so the configured-but-unpublished gut and liver tracker atlases emit neither route — verified absent from `out/`.
- **The `buildStaticPaths` refactor changes nothing**: diffed the full emitted page list before and after, 63 → 63 pages, zero difference.

## Notes for review

- **Pre-existing bug found during review, filed as #3203, not fixed here.** `getPublishedAtlases` nulls its cached promise inside the `.catch()`, so a transient `/api/published-atlases` failure makes the publication gate non-deterministic *within one build* — one caller sees `[]`, the next re-fetches and may succeed. Two page modules consult that gate, so a build can half-generate a tracker atlas (an overview shipping a Source Datasets tab pointing at a page that was never written) and still exit 0. **Not a regression from this PR**: before the split, `index.tsx` and `datasets.tsx` both used the gated `getStaticPaths`, so two independent gated calls already existed; this PR keeps the count at two. Fixing it means changing the tracker layer's failure semantics, which doesn't belong in a routing change.
- **The old tracker `/datasets` URL 404s**, as #3198 and #3193 both specify. Exposure is exactly one URL — `/hca-bio-networks/breast/atlases/breast-v1/datasets` — since Breast v1.0 is the only published tracker atlas; gut and liver emit nothing today. It has existed only since 2026-08-21 (#3150), is reached through in-app tab navigation, and is not linked externally.
- **Deliberately not renamed:** `Datasets/` → `SourceStudies/`. It would be symmetric with the new sibling, but the route is still `/datasets`, and reusing the `SourceStudies/` name #3194 just deleted for a different, Azul-sourced table would muddy the history.
- **Second pre-existing find, filed as #3205, not fixed here.** The atlas overview still routes through `getTrackerContentStaticProps`, which fetches component-atlases, source-datasets *and* source-studies for every tracker atlas — but the Overview components read only `atlas.integratedAtlases`. Two of three build-time requests are dead on that route, both in an unguarded `Promise.all`, and the joined `trackerSourceDatasets` records still ship in the overview's `__NEXT_DATA__`. Same shape as the cleanup in #3197. Pre-existing, and #3198 is what makes each route single-purpose enough to fix it cleanly.


NoopDog (approved):
Reviewed the route split, the component move, and the `buildStaticPaths` refactor. No correctness bugs found.

Verified independently:

- The rewritten relative import depths in the moved `SourceDatasets/` tree are all correct; `tsc --noEmit` and `lint` pass clean on the branch (pre-existing warnings only).
- `buildStaticPaths` is behaviour-preserving. `getStaticPaths`'s predicate `atlas.tracker ? isPublishedTrackerAtlas(atlas) : true` is exactly the old `if (atlas.tracker && !published) continue`, and the helper's early `if (!atlas.tracker) return false` covers the tracker-only variant.
- The `nonTrackerTabName` removal is a no-op for output — `/datasets` used to receive "Source Datasets" and remap it; it now receives "Source Studies" directly. Same title.
- No orphaned references: the old `Datasets/.../tracker/` tree is gone, `AtlasContext.trackerSourceDatasets` is optional so dropping it from `datasets.tsx`'s provider is safe, and nothing outside `tabs.tsx` links to either sub-route.
- Overview and tab target can't diverge — `index.tsx` and `source-datasets.tsx` both gate on `isPublishedTrackerAtlas`.
- The #3203 analysis checks out. `getPublishedAtlases` does null its cached promise in `.catch()`, and the gated-caller count is genuinely unchanged at two. `getNonTrackerStaticPaths` no longer touches the gate at all, which slightly reduces the exposed surface. Agreed it doesn't belong here.

Accepting the deliberate 404 on `/hca-bio-networks/breast/atlases/breast-v1/datasets`. `output: "export"` gives us no redirect mechanism, and the exposure is scoped as you describe — one URL, in-app navigation only, no external links (confirmed by grep). Not worth a CDN rule.

Two non-blocking nits:

- `SourceDatasets/components/MainColumn/sourceDatasets.tsx:14` — stale JSDoc still says "Tracker source datasets view for the Datasets tab". Post-split that reads as the unrelated non-tracker Azul route.
- `pages/.../source-datasets.tsx:11` — `import * as atlasPages` alongside a named import from the same module. Mirrors the existing pages, so fine to leave.

Thanks for flagging the path move for #3200/#3195/#3196/#3199 up front.

Issue #3195, opened 2026-09-08
feat: atlas source datasets table: stack file name over source study in one pinned column


NoopDog:
Part of #3193.

## Problem

The tracker source datasets table opens with two columns that between them say one thing badly:

- **Dataset** (pinned, first) shows the tracker's `title` — a citation-style label, e.g. "Gray et al. 2022, Developmental Cell".
- **Source Study** shows `publicationString` — e.g. "Gray (2022) Developmental Cell", linked to the DOI.

They restate each other, they burn two columns of a wide table, and neither identifies the file the Download button actually hands you.

Replace both with a single pinned column that stacks the **file name** over the **source study**, following the stacked-cell pattern used in BRC Analytics.

## Depends on

#3200 — it exports the shared `buildVersionedFileName` helper and puts the versioned file name on `DatasetAsset`. This column consumes it; it must not derive the `-r<revision>` suffix itself.

## Approach — which name to show

Show the **revisioned** published file name, including extension — the name of the file the row's Download button actually delivers:

> `ihbca-v1-gray-2022-r1.h5ad`

Both parts come from the tracker's source-dataset record: `baseFileName` (`ihbca-v1-gray-2022.h5ad`) and `revision` (`1`), combined by `buildVersionedFileName` as `<stem>-r<revision><ext>`. After #3200 that value is already on the row's `datasetAsset` — read it from there.

For reference, the record's three name-ish fields, verified against the live tracker for Breast v1.0 (Gray et al.):

| Field | Value | Use |
| --- | --- | --- |
| `baseFileName` | `ihbca-v1-gray-2022.h5ad` | Input to the versioned name — not shown as-is |
| `downloadName` | `ihbca-v1-gray-2022` | Stem only, unrevisioned; not on `TrackerSourceDataset` |
| `fileName` | `gray2022-r1-wip-2-edit-2026-08-25-03-50-06.h5ad` | Tracker-internal working name — must not be shown |

Note the tracker's own `fileName` field is *not* the file name to display, despite the name.

## Approach — the stacked cell

[...]

Issue #3190, opened 2026-09-08
test: add Jest unit test capability, modelled on hca-atlas-tracker


NoopDog:
Part of #3189.

## Problem

The repo has no unit/integration test capability: no Jest (or any runner) in `devDependencies`, no `test` script in `package.json`, and no test directories. Pure logic — view builders, `utils/`, hooks, mappers — cannot be tested at all today.

The Clever Canary standard requires CI to run "lint, typecheck, **and tests**" on every pull request (`cc/STANDARDS.md`, conformance check A7). Lint, format, and typecheck already run; tests do not exist.

## Approach

Port the setup from [`clevercanary/hca-atlas-tracker`](https://github.com/clevercanary/hca-atlas-tracker), the sibling Next.js + TypeScript app, rather than inventing one. Its `jest.config.js`:

- Jest 30 wired through `next/jest` (`createJestConfig`), so `next.config.mjs` and env files load in tests
- `testEnvironment: "jest-environment-jsdom"`
- `moduleNameMapper` for the `@/` path alias
- `setupFilesAfterEnv: ["<rootDir>/testing/setup.ts"]`
- `moduleDirectories: ["node_modules"]` (no `<rootDir>`)
- `testPathIgnorePatterns: ["/node_modules/", "/e2e/"]` — keeps Playwright specs out of the Jest run

## Definition of done

- [ ] `jest`, `jest-environment-jsdom`, `@testing-library/react`, `@testing-library/jest-dom` added to `devDependencies`
- [ ] `jest.config.js` added, adapted from the tracker's (this repo has no `@/` alias and no DB, so the mapper and any DB setup drop out)
- [ ] `testing/setup.ts` added for shared setup
- [ ] `"test": "jest"` script added to `package.json` (the tracker uses `--runInBand` for DB serialization; not needed here)
- [ ] `__tests__/` directory established
- [ ] Seed tests written for `utils/trackerNetwork.ts` — `buildVersionedFileName` and `splitFileName`, covering the `<stem>-r<revision><ext>` convention, names with no extension, and names with multiple dots
- [ ] `coverage/` added to `.gitignore` and the ESLint ignore list

Wiring `npm test` into the CI workflow is tracked separately in the CI sub-issue, so this issue can land without touching `.github/workflows/`.

## Out of scope

Backfilling tests across the codebase, and fixing the duplicated `-r<revision>` literal in `viewBuilder.ts:41`. This issue stands up the capability and seeds it.

Issue #3189, opened 2026-09-08
[Epic] Testing capability and CI conformance


NoopDog:
## Summary

The Data Portal has **no test capability of any kind** — no test runner in `devDependencies`, no `test` script, no test directories, and no test step in CI. This epic tracks standing that up, and closing the remaining gaps between what CI runs today and what the Clever Canary repo standards require.

This is an **epic**. Each piece of work is tracked in a sub-issue linked below.

## What CI runs today

`.github/workflows/run-checks.yml` already runs, on every pull request:

- Next.js production build (`npm run build-prod:data-portal`)
- Prettier (`npm run check-format`)
- ESLint (`npm run lint`)
- Type check (`npx tsc --noEmit`)

So linting, formatting, and type checking are **already covered** and are not part of this epic.

## What is missing

| Gap | Standard | Sub-issue |
| --- | --- | --- |
| No unit/integration tests (Jest) | A7 — CI runs lint, typecheck, **and tests** | #3190 |
| No end-to-end tests (Playwright) | A10 — web apps MUST have Playwright e2e in CI | #3188 |
| Test jobs not wired into CI, plus workflow hygiene | A7 | #3191 |
| No security workflow at all | A8 — static analysis, dependency review, gitleaks | #3192 |
| No gitleaks pre-commit hook | A17 | #3192 |

## Reference implementation

[`clevercanary/hca-atlas-tracker`](https://github.com/clevercanary/hca-atlas-tracker) is the sibling Next.js + TypeScript app and the model to follow for Jest. It has:

- `jest.config.js` — Jest 30 via `next/jest`, `jest-environment-jsdom`, `moduleNameMapper` for the `@/` alias, `setupFilesAfterEnv: testing/setup.ts`
- `@testing-library/react` and `@testing-library/jest-dom`
- ~100 test files in `__tests__/`, with shared helpers in `testing/`
- `"test": "jest --runInBand"`, run in CI as a separate `test` job

Note the tracker has `@playwright/test` in `devDependencies` but no Playwright config and no specs, so it is **not** a reference for the Playwright work — that gap is being tracked in the tracker repo separately.

## What motivated this

While reviewing tracker download filenames, two pure functions in `utils/trackerNetwork.ts` turned out to have no coverage and no way to add any:

- `buildVersionedFileName` (`utils/trackerNetwork.ts:163-169`) — builds the versioned filename `<stem>-r<revision><ext>`
- `splitFileName` — splits stem from extension

[...]

PR #3206, opened 2026-09-09
refactor: derive the tracker -r<revision> file name from one shared helper (#3200)

### LINKED ISSUE #3200: refactor: derive the tracker -r<revision> file name from one shared helper

frano-m:
## Summary

Closes #3200. Part of #3193.

The `<stem>-r<revision><ext>` convention was derived three different ways, in two different styles. This carries the published name on the dataset asset — built once, alongside the download URL it already derives — so the convention lives in exactly one place. `#3195` was about to add a fourth derivation.

**Before:**

| Site | How it derived the name |
| --- | --- |
| `utils/trackerNetwork.ts` | `buildVersionedFileName` — canonical, but private |
| source datasets download cell | re-implemented the suffix inline as `` `${stem}-r${revision}` `` |
| integrated objects download cell | popped the last segment off `asset.downloadURL` and split *that* |

**After:** `buildVersionedFileName` is the only place `-r${revision}` appears, and nothing derives a file name from a URL.

## Changes

- `@types/network.ts` — `DatasetAsset.versionedFileName?`, the published name with extension. Optional, since CXG assets are built elsewhere and rendered by `CXGDownloadCell`, which does not use it.
- `utils/trackerNetwork.ts` — exports `buildVersionedFileName`; `buildTrackerDatasetAsset` derives the name once and passes it to `buildTrackerDownloadUrl` (whose signature and JSDoc changed to match).
- Both download cells read the name off the asset via a new shared `buildTrackerDownloadCellProps`.

## ⚠️ Deviation from the issue: the field is `versionedFileName`, not `fileName`

#3200 specifies adding **`fileName`** to `DatasetAsset`. I named it **`versionedFileName`** instead, because `fileName` would have collided with an existing field of the same name meaning something else.

`TrackerSourceDataset.fileName` and `TrackerComponentAtlas.fileName` already exist as required strings holding the tracker's **internal working name**. Checked against all seven live Breast v1.0 records — they differ from the published name on *every one*:

```
tracker API fileName            gray2022-r1-wip-2-edit-2026-08-25-03-50-06.h5ad
baseFileName + revision         ihbca-v1-gray-2022-r1.h5ad     <- what S3 actually serves
```

Had the asset field also been called `fileName`, then `sd.fileName !== sd.datasetAsset.fileName` on every record, with the wrong one being the shorter, more obvious reach. #3195's File Name column is exactly where that mistake would be made, and it would be silent — the working name is a required `string` that renders as a plausible `.h5ad`.

So alongside the rename, both tracker `fileName` fields now carry a comment saying they are the working name and pointing at `versionedFileName`. I've commented on #3195 with the same warning.

The issue's *intent* — the asset carries the versioned name, derived in one place — is met. Say the word if you'd rather it match the issue's literal wording.

## Also addressed from review

- **Shared cell props.** The two download cells had become structurally identical (guard, `splitFileName`, same four props). `buildTrackerDownloadCellProps` puts the stem/extension display convention in one place too, so a change to the naming convention can't drift between them. Both call sites are now three lines.
- **URL encoding.** `buildTrackerDownloadUrl` now wraps the name in `encodeURIComponent`. Pre-existing gap — a name containing a space, `#`, or `?` produced a broken link. Every generated URL is unchanged today, since all current names are slugs.

## Deliberately not done

- **No tracker-specific `DatasetAsset` subtype.** Suggested so the render paths could require `versionedFileName` at compile time rather than degrading to a hidden download button. `buildTrackerDatasetAsset` is the only tracker asset builder and always sets it, and #3200 specifies the field as optional. Reasonable follow-up, not this PR.
- **No alphabetical reorder of `utils/trackerNetwork.ts`.** It is already non-alphabetical at HEAD (`mapTrackerComponentAtlasToIntegratedAtlas` precedes `buildTrackerDatasetAsset`), so moving one function wouldn't make it compliant — it needs its own pass.
- **`buildVersionedFileName` is exported but has no external caller yet.** That is deliberate and specified: #3189 cites this function and `splitFileName` as untestable today, and #3200's notes say to seed unit tests for exactly those two once #3190 lands.

## Verification

- **All 175 `.h5ad` download URLs byte-for-byte identical** to a baseline captured from the build before any change — including after adding `encodeURIComponent`.
- **Displayed names identical** on `source-datasets.html` and `breast-v1.html` (8 each).
- Gray et al. resolves to `versionedFileName: "ihbca-v1-gray-2022-r1.h5ad"`, rendered as stem `ihbca-v1-gray-2022-r1` plus format `.h5ad` — the AC's exact case. All 7 assets carry it.
- `` `-r${revision}` `` appears once, at `utils/trackerNetwork.ts:191`. `grep` for `downloadURL.split` returns nothing.
- `npm run lint` (0 errors), `npm run check-format`, `npx tsc --noEmit`, `npm run build-prod:data-portal` (64/64) all pass — re-run after rebasing onto `main` post-#3204.

## Tests

Skipped: #3190 has not landed, so there is still no test tooling in the repo (no `test` script, no Jest/Vitest in `devDependencies`, no config). #3200's note to seed `buildVersionedFileName` / `splitFileName` tests carries over to whoever lands #3190 — the export is already in place for it.

### DIFF
--- a/@types/network.ts
+++ b/@types/network.ts
@@ -113,9 +113,13 @@ export type IntegratedAtlasRow = AtlasRow &
 
 export interface DatasetAsset {
   downloadURL: string;
-  fileName?: string; // Versioned file name with extension, e.g. "ihbca-v1-gray-2022-r1.h5ad". Set for tracker-built assets only; CXG assets are rendered by CXGDownloadCell, which does not use it.
   fileSize: number;
   fileType: CXG_DATASET_FILE_TYPE;
+  // Published file name with extension, e.g. "ihbca-v1-gray-2022-r1.h5ad",
+  // derived by `buildVersionedFileName`. NOT the tracker's own `fileName`,
+  // which is an internal working name. Set for tracker-built assets only; CXG
+  // assets are rendered by CXGDownloadCell, which does not use it.
+  versionedFileName?: string;
 }
 
 export interface IntegratedAtlas {
@@ -201,6 +205,10 @@ export interface TrackerComponentAtlas {
   cellCount: number;
   disease: string[];
   fileId: string;
+  // The tracker's internal working file name, e.g.
+  // "gray2022-r1-wip-2-edit-2026-08-25-03-50-06.h5ad". This is NOT the
+  // published object name - use `datasetAsset.versionedFileName` for anything
+  // user-facing or for building a download URL.
   fileName: string;
   geneCount: number;
   id: string;
@@ -227,6 +235,10 @@ export interface TrackerSourceDataset {
   disease: string[];
   doi: string | null;
   fileId: string;
+  // The tracker's internal working file name, e.g.
+  // "gray2022-r1-wip-2-edit-2026-08-25-03-50-06.h5ad". This is NOT the
+  // published object name - use `datasetAsset.versionedFileName` for anything
+  // user-facing or for building a download URL.
   fileName: string;
   geneCount: number;
   hcaProjectId: string | null; // Joined from the source study; null when the study has no HCA project or the join misses.
--- a/@types/network.ts
+++ b/@types/network.ts
@@ -113,6 +113,7 @@ export type IntegratedAtlasRow = AtlasRow &
 
 export interface DatasetAsset {
   downloadURL: string;
+  fileName?: string; // Versioned file name with extension, e.g. "ihbca-v1-gray-2022-r1.h5ad". Set for tracker-built assets only; CXG assets are rendered by CXGDownloadCell, which does not use it.
   fileSize: number;
   fileType: CXG_DATASET_FILE_TYPE;
 }
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.ts
@@ -8,7 +8,7 @@ import * as C from "../../../../../../../../..";
 import type { TrackerSourceDataset } from "../../../../../../../../../../@types/network";
 import {
   buildTrackerAnalysisPortals,
-  splitFileName,
+  buildTrackerDownloadCellProps,
 } from "../../../../../../../../../../utils/trackerNetwork";
 
 const DOI_BASE_URL = "https://doi.org/";
@@ -33,15 +33,9 @@ export function renderCellCount(
 export function renderDownload(
   ctx: CellContext<TrackerSourceDataset, unknown>
 ): JSX.Element | null {
-  const { datasetAsset } = ctx.row.original;
-  if (!datasetAsset?.fileName) return null;
-  const { ext: format, stem } = splitFileName(datasetAsset.fileName);
-  return C.TrackerDownloadCell({
-    downloadUrl: datasetAsset.downloadURL,
-    fileName: stem,
-    fileSize: datasetAsset.fileSize,
-    format,
-  });
+  const props = buildTrackerDownloadCellProps(ctx.row.original.datasetAsset);
+  if (!props) return null;
+  return C.TrackerDownloadCell(props);
 }
 
 /**
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.ts
@@ -33,12 +33,12 @@ export function renderCellCount(
 export function renderDownload(
   ctx: CellContext<TrackerSourceDataset, unknown>
 ): JSX.Element | null {
-  const { baseFileName, datasetAsset, revision } = ctx.row.original;
-  if (!datasetAsset) return null;
-  const { ext: format, stem } = splitFileName(baseFileName);
+  const { datasetAsset } = ctx.row.original;
+  if (!datasetAsset?.fileName) return null;
+  const { ext: format, stem } = splitFileName(datasetAsset.fileName);
   return C.TrackerDownloadCell({
     downloadUrl: datasetAsset.downloadURL,
-    fileName: `${stem}-r${revision}`,
+    fileName: stem,
     fileSize: datasetAsset.fileSize,
     format,
   });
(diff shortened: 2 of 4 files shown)


NoopDog (approved):
Approving. Pulled the branch and checked it against `main`.

**Verified the core claims:**

- `` `-r${revision}` `` appears exactly once on the branch (`utils/trackerNetwork.ts:190`); `downloadURL.split` returns nothing.
- No missed call site. I initially flagged `components/.../Datasets/.../MainColumn/tracker/components/table/viewBuilder.ts` as a fourth inline derivation — it still carries the `` `${stem}-r${revision}` `` copy at `main`. That file no longer exists on this branch; #3204 moved it to `SourceDatasets/`, which is the one you edit here. False alarm.
- Both cell paths produce identical output. The old source-datasets path was `splitFileName(baseFileName).stem + "-r" + revision`; the new one is `splitFileName(buildVersionedFileName(base, rev)).stem` — same string. The old integrated path parsed the URL's last segment, which was that same name.
- `encodeURIComponent` is applied only inside `buildTrackerDownloadUrl`, so the asset keeps the raw name for display and the encoded form for the href. That's the right split, and a no-op for every slug-shaped name today.

**On the `versionedFileName` deviation:** endorsed, keep it. The evidence is concrete — `TrackerSourceDataset.fileName` is a required string holding `gray2022-r1-wip-2-edit-....h5ad` while S3 serves `ihbca-v1-gray-2022-r1.h5ad`, differing on all seven records — and a same-named field would make the wrong one the shorter reach, rendering as a plausible `.h5ad` with no error. The two warning comments on the tracker `fileName` fields are the right mitigation. #3200's intent is met; the literal wording isn't worth the trap.

**Notes, none blocking:**

1. **The props shape is now declared in two places** (`utils/trackerNetwork.ts:145-149`). `buildTrackerDownloadCellProps` returns an inline `{ downloadUrl, fileName, fileSize, format }` that structurally duplicates `TrackerDownloadCellProps` in `components/common/Table/components/Cell/components/TrackerDownloadCell/types.ts`. For a PR whose thesis is "the convention lives in exactly one place," that's the one loose thread. `import type { TrackerDownloadCellProps }` is zero-runtime; the tradeoff is that no `utils/` module imports from `components/` today, so it introduces a new (type-only) direction. Your call either way — worth a line in the PR if you decide against it.

2. **`buildTrackerDownloadCellProps` is a view-model concern living in `utils/`.** Cell props are built in `viewModelBuilders/` and the table `viewBuilder.ts` files everywhere else. Defensible here, since both consumers already import this module and the point is to sit next to `splitFileName` — just noting it's the structural odd one out.

3. **Two exports with zero callers.** With both viewBuilders switched over, `splitFileName` no longer has a consumer outside its own module, so it and the newly-exported `buildVersionedFileName` are both exported for tests that don't exist yet. That's your stated intent and I'd keep it, but nothing enforces the follow-up and lint won't flag it. If #3190 slips, consider making `splitFileName` private again.

4. **Silent degrade on a missing name.** `versionedFileName?` optional plus `buildTrackerDownloadCellProps` returning `null` means an asset without the field renders no download button at all, with no signal. Agreed the tracker-specific subtype isn't this PR — just make sure it's actually ticketed rather than only living in the PR description.

Issue #3205, opened 2026-09-09
refactor: the atlas overview page fetches source datasets and source studies it never renders


frano-m:
## Problem

`getTrackerContentStaticProps` in `utils/trackerAtlasPages.ts` is the single static-props builder for every tracker-sourced route, and it unconditionally fetches all three tracker collections:

```ts
const [componentAtlases, sourceDatasets, sourceStudies] = await Promise.all([
  fetchTrackerComponentAtlases(atlasId),
  fetchTrackerSourceDatasets(atlasId),
  fetchTrackerSourceStudies(atlasId),
]);
```

Its only per-page input is `tabName`, used solely for the page title, so it has no notion of what the page being built actually needs. The atlas overview therefore inherits the source-datasets page's data requirements.

The Overview components read only `atlas.integratedAtlases` (`Overview/components/MainColumn/mainColumn.tsx:27`) plus `atlas` and `network`. Nothing under `Overview/` touches `trackerSourceDatasets`. `capId` and `cxgDataPortal` come from `resolveTrackerAtlas`, not from the three fetches — so of the three, only `fetchTrackerComponentAtlases` feeds the overview at all.

## Cost

Per tracker atlas, on the overview route:

- **Two wasted build-time HTTP requests** — source-datasets and source-studies. Both sit in an unguarded `Promise.all`, so a 5xx or timeout from either fails the static export of the overview page for data it does not render.
- **Wasted payload** — the joined `trackerSourceDatasets` records still serialize into the overview's `__NEXT_DATA__`, because `index.tsx` destructures `trackerSourceDatasets` and threads it into `AtlasProvider`. For Breast v1.0 that is 7 records carrying `journal`, `referenceAuthor` and `hcaProjectId` from #3197's join, shipped to every visitor of a page that never reads them.

This is the same shape as the cleanup in #3197, which removed the readerless `trackerSourceStudies` prop and saved ~36KB per page.

## Why it exists

Historical rather than intentional. When the tracker view was built (#3150) it had three tabs — overview, source datasets, source studies — all sharing `getContentStaticProps`, so one props builder serving them all was the simple choice. #3194 removed the source studies tab and #3198 split the route so each page now serves exactly one kind of content, which leaves the sharing as the only reason the overview over-fetches.

## Approach

Make the builder aware of what the route needs, rather than fetching everything. Options worth weighing while implementing:

[...]

PR #3208, opened 2026-09-09
feat: atlas source datasets table: stack file name over source study in one pinned column (#3195)

### LINKED ISSUE #3195: feat: atlas source datasets table: stack file name over source study in one pinned column

frano-m:
## Summary

Closes #3195. Part of #3193.

The table opened with two columns saying one thing badly — **Dataset** (`title`, e.g. "Gray et al. 2022, Developmental Cell") and **Source Study** (`publicationString`, e.g. "Gray (2022) Developmental Cell") — which restated each other, burned two columns, and identified nothing about the file the Download button delivers. Both are replaced by one pinned column stacking the published file name over the DOI-linked citation.

## Changes

Under `SourceDatasets/components/MainColumn/components/table/`:

- **New `components/FileNameCell/`** — `fileNameCell.tsx`, `types.ts`, `constants.ts`. A MUI `Stack` with the file name at `body-400` over the citation at `ink.light` / `body-small-400`, matching the BRC Analytics stacked-cell treatment.
- `accessor.ts` — `buildVersionedFileNameValue`, so the column sorts on the value it displays rather than the unrevisioned base name.
- `columns.ts` — `TITLE` becomes `FILE_NAME` (`id: "fileName"`, header **File Name**, pinned, first). `SOURCE_STUDY` keeps its `accessorKey` and filter config but loses its `cell`.
- `utils.ts` — `sourceStudy: false` joins the filter-only columns.
- `hook.ts` — initial sort moves from the removed `"title"` to `"fileName"`.

The name comes from `datasetAsset.versionedFileName` (#3200), so the `-r<revision>` suffix is not derived here.

## ⚠️ Deviation: `renderSourceStudy` was removed, not reused

#3195 says to "reuse the existing `renderSourceStudy` link logic from `viewBuilder.ts`". I moved that logic into the new cell and **deleted the function**, rather than keeping it.

Making `SOURCE_STUDY` filter-only means its `cell` is never reached — hidden columns are not rendered — so `renderSourceStudy` would have been dead code. The repo's other filter-only columns (`INTEGRATED_OBJECTS`, `JOURNAL`, `REFERENCE_AUTHOR`) carry no `cell` at all, so I matched that shape and also dropped the now-meaningless `meta.width`. `DOI_BASE_URL` moved from `viewBuilder.ts` into the new cell's `constants.ts`, its only remaining consumer.

I read "reuse the logic" as reuse-the-behaviour rather than keep-the-function; flag it if you would rather the function stayed.

## Verification

Against the built page for Breast v1.0:

- Headers are now `File Name, Assay, Tissue, Disease, Cell Count, Explore, Download` — both `Dataset` and the `Source Study` column are gone.
- **The Source Study filter survives** — all seven chips still render in the sidebar, faceting over `publicationString`.
- All seven versioned names render **as text in the document body**, not merely in `__NEXT_DATA__`, each with a `doi.org` link carrying `rel="noopener noreferrer"`.
- **No tracker working name is shown**: zero occurrences of `-wip-` or `edit-2026` in the rendered body. The seven hits inside `__NEXT_DATA__` are the tracker's own `fileName` field, which the page carries but never displays.
- Loads sorted ascending by file name (gray → twigger), checked against `sorted()`.
- No `title` column id or `row.title` reference remains in the table directory.
- Column width verified in the browser at desktop width.

## From review

- **The pinned cell can no longer render empty.** `datasetAsset` is optional on the type and `publicationString` is nullable, so a row with neither would have produced a bare `Stack` — and because this is the pinned column it identifies the row in the collapsed layout. The primary line now falls back to `baseFileName` (a required field), and `buildVersionedFileNameValue` mirrors that fallback so such rows cannot sort on an empty string either. **Not** falling back to `title`, which #3195 puts out of scope for the pinned slot.
- **`buildTrackerDownloadCellProps` now returns the exported `TrackerDownloadCellProps`** instead of an inline duplicate of that shape — this also answers a note on #3206.
- `FILE_NAME` moved to its alphabetical position among the column consts; `JSX` is a `type` import.
- **Cells now render as real elements.** `viewBuilder.ts` is renamed `viewBuilder.tsx` and all three renderers return JSX (`<TrackerDownloadCell {...props} />`, `<AnalysisPortalCell … />`, `<FileNameCell … />`) instead of invoking components as plain functions, so each cell gets its own component boundary. This mattered beyond the new cell: `renderDownload` was calling `TrackerDownloadCell`, which uses `useDialog()`, so that hook was attaching to whichever component rendered the cell. The `import * as C` barrel namespace is replaced by direct imports. The same pattern remains at 17 call sites in `viewModelBuilders.ts` (including a second `useDialog` component) — out of scope here, filed as #3211.

## Not addressed

- **Five tracker source-dataset fields now ship in props with no reader** — `fileName`, `fileId`, `sourceStudyTitle`, `geneCount`, and `title`, which this PR orphans by removing its only consumer. Added to #3205 rather than fixed here, since trimming them needs a projection step between the API response and the props.

## Out of scope

The external-link affordance on the citation line — #3196, which builds on this cell.

image

### DIFF
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/accessor.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/accessor.ts
@@ -36,3 +36,14 @@ export function buildJournal(row: TrackerSourceDataset): string {
 export function buildReferenceAuthor(row: TrackerSourceDataset): string {
   return row.referenceAuthor || LABEL.UNSPECIFIED;
 }
+
+/**
+ * Returns the file name shown in the pinned column, so the column sorts on the
+ * value it displays rather than on the unrevisioned base name. Mirrors the
+ * fallback in `FileNameCell` so no row ever sorts on an empty value.
+ * @param row - Tracker source dataset.
+ * @returns versioned file name, falling back to the base file name.
+ */
+export function buildVersionedFileNameValue(row: TrackerSourceDataset): string {
+  return row.datasetAsset?.versionedFileName || row.baseFileName;
+}
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
@@ -1,0 +1,48 @@
+import {
+  ANCHOR_TARGET,
+  REL_ATTRIBUTE,
+} from "@databiosphere/findable-ui/lib/components/Links/common/entities";
+import { Link } from "@databiosphere/findable-ui/lib/components/Links/components/Link/link";
+import { TYPOGRAPHY_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/typography";
+import { Stack, Typography } from "@mui/material";
+import type { JSX } from "react";
+import { DOI_BASE_URL } from "./constants";
+import type { Props } from "./types";
+
+/**
+ * Pinned cell stacking the published file name over the source study citation.
+ * The primary line is the versioned name carried on the dataset asset, so it
+ * always matches the file the row's Download button delivers.
+ * @param props - Component props.
+ * @param props.row - Tracker source dataset.
+ * @returns stacked file name and source study cell.
+ */
+export const FileNameCell = ({ row }: Props): JSX.Element => {
+  const { baseFileName, datasetAsset, doi, publicationString } = row;
+  // `versionedFileName` is set for every tracker source dataset, but it is
+  // optional on the shared asset type - fall back to the unversioned base name
+  // so the pinned column, which identifies the row in the collapsed layout,
+  // can never render empty.
+  const fileName = datasetAsset?.versionedFileName || baseFileName;
+  return (
+    <Stack spacing={2} useFlexGap>
+      <Typography variant={TYPOGRAPHY_PROPS.VARIANT.BODY_400}>
+        {fileName}
+      </Typography>
+      {publicationString && (
+        <Typography
+          color={TYPOGRAPHY_PROPS.COLOR.INK_LIGHT}
+          component="div"
+          variant={TYPOGRAPHY_PROPS.VARIANT.BODY_SMALL_400}
+        >
+          <Link
+            label={publicationString}
+            rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
+            target={ANCHOR_TARGET.BLANK}
+            url={doi ? `${DOI_BASE_URL}${doi}` : ""}
+          />
+        </Typography>
+      )}
+    </Stack>
+  );
+};
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
@@ -1,11 +1,12 @@
 import type { CellContext } from "@tanstack/react-table";
 import type { JSX } from "react";
-import * as C from "../../../../../../../../..";
 import type { TrackerSourceDataset } from "../../../../../../../../../../@types/network";
 import {
   buildTrackerAnalysisPortals,
   buildTrackerDownloadCellProps,
 } from "../../../../../../../../../../utils/trackerNetwork";
+import { TrackerDownloadCell } from "../../../../../../../../../common/Table/components/Cell/components/TrackerDownloadCell/trackerDownloadCell";
+import { AnalysisPortalCell } from "../../../../../Overview/components/MainColumn/components/AnalysisPortalCell/analysisPortalCell";
 import { FileNameCell } from "./components/FileNameCell/fileNameCell";
 
 /**
@@ -30,7 +31,7 @@ export function renderDownload(
 ): JSX.Element | null {
   const props = buildTrackerDownloadCellProps(ctx.row.original.datasetAsset);
   if (!props) return null;
-  return C.TrackerDownloadCell(props);
+  return <TrackerDownloadCell {...props} />;
 }
 
 /**
@@ -43,7 +44,7 @@ export function renderExplore(
 ): JSX.Element | null {
   const analysisPortals = buildTrackerAnalysisPortals(ctx.row.original.capUrl);
   if (analysisPortals.length === 0) return null;
-  return C.AnalysisPortalCell({ analysisPortals });
+  return <AnalysisPortalCell analysisPortals={analysisPortals} />;
 }
 
 /**
@@ -55,5 +56,5 @@ export function renderExplore(
 export function renderFileName(
   ctx: CellContext<TrackerSourceDataset, unknown>
 ): JSX.Element {
-  return FileNameCell({ row: ctx.row.original });
+  return <FileNameCell row={ctx.row.original} />;
 }
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/columns.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/columns.ts
@@ -14,12 +14,13 @@ import {
   buildIntegratedObjects,
   buildJournal,
   buildReferenceAuthor,
+  buildVersionedFileNameValue,
 } from "./accessor";
 import {
   renderCellCount,
   renderDownload,
   renderExplore,
-  renderSourceStudy,
+  renderFileName,
 } from "./viewBuilder";
 
 const ASSAY = {
@@ -76,6 +77,16 @@ const EXPLORE = {
   meta: { width: "auto" },
 } as ColumnDef<TrackerSourceDataset>;
 
+const FILE_NAME = {
+  accessorFn: buildVersionedFileNameValue,
+  cell: renderFileName,
+  enableColumnFilter: false,
+  header: "File Name",
+  id: "fileName",
+  meta: { columnPinned: true, width: { max: "2.4fr", min: "260px" } },
+  sortingFn,
+} as ColumnDef<TrackerSourceDataset>;
+
 const INTEGRATED_OBJECTS = {
   accessorFn: buildIntegratedObjects,
   enableColumnFilter: true,
@@ -102,13 +113,10 @@ const REFERENCE_AUTHOR = {
 
 const SOURCE_STUDY = {
   accessorKey: "publicationString",
-  cell: renderSourceStudy,
   enableColumnFilter: true,
   filterFn: "arrIncludesSome",
   header: "Source Study",
   id: "sourceStudy",
-  meta: { width: { max: "1.2fr", min: "200px" } },
-  sortingFn,
 } as ColumnDef<TrackerSourceDataset>;
 
 const TISSUE = {
@@ -122,18 +130,8 @@ const TISSUE = {
   sortingFn,
 } as ColumnDef<TrackerSourceDataset>;
 
-const TITLE = {
-  accessorKey: "title",
-  enableColumnFilter: false,
-  header: "Dataset",
-  id: "title",
-  meta: { columnPinned: true, width: { max: "2fr", min: "200px" } },
-  sortingFn,
-} as ColumnDef<TrackerSourceDataset>;
-
 export const COLUMNS: ColumnDef<TrackerSourceDataset>[] = [
-  TITLE,
-  SOURCE_STUDY,
+  FILE_NAME,
   ASSAY,
   TISSUE,
   DISEASE,
@@ -143,4 +141,5 @@ export const COLUMNS: ColumnDef<TrackerSourceDataset>[] = [
   INTEGRATED_OBJECTS,
   JOURNAL,
   REFERENCE_AUTHOR,
+  SOURCE_STUDY,
 ];
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/constants.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/constants.ts
@@ -1,0 +1,1 @@
+export const DOI_BASE_URL = "https://doi.org/";
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/types.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/types.ts
@@ -1,0 +1,5 @@
+import type { TrackerSourceDataset } from "../../../../../../../../../../../../@types/network";
+
+export interface Props {
+  row: TrackerSourceDataset;
+}
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
@@ -28,7 +28,7 @@ export const useTable = (
     enableRowPreview: false,
     getRowId: (row) => row.id,
     initialState: {
-      sorting: [{ desc: SORT_DIRECTION.ASCENDING, id: "title" }],
+      sorting: [{ desc: SORT_DIRECTION.ASCENDING, id: "fileName" }],
     },
     meta: META,
     state: {
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/utils.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/utils.ts
@@ -3,9 +3,9 @@ import type { TrackerSourceDataset } from "../../../../../../../../../../@types/
 
 /**
  * Returns the column visibility state for the tracker source datasets table.
- * The "Integrated Object", "Journal" and "Reference Author" columns are
- * filter-only, and the "Explore" column is displayed only when at least one
- * dataset has a CAP link.
+ * The "Integrated Object", "Journal", "Reference Author" and "Source Study"
+ * columns are filter-only, and the "Explore" column is displayed only when at
+ * least one dataset has a CAP link.
  * @param data - Tracker source datasets.
  * @returns column visibility state.
  */
@@ -17,5 +17,6 @@ export function getColumnVisibility(
     integratedObject: false,
     journal: false,
     referenceAuthor: false,
+    sourceStudy: false,
   };
 }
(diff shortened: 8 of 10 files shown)


components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
```diff
@@ -0,0 +1,48 @@
...
+          variant={TYPOGRAPHY_PROPS.VARIANT.BODY_SMALL_400}
+        >
+          <Link
+            label={publicationString}
+            rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
+            target={ANCHOR_TARGET.BLANK}
+            url={doi ? `${DOI_BASE_URL}${doi}` : ""}
+          />
+        </Typography>
+      )}
```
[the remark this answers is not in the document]
frano-m: No empty-href anchor is produced here — findable-ui's `Link` already degrades to plain text for an invalid URL, which is the behaviour you're asking for.

Traced through `lib/components/Links/components/Link/link.js` with `url=""`:

- `isURLObjectWithHrefAndQuery("")` → false (it's a string)
- `isURLString("")` → **true** (`typeof value === "string"`)
- `isClientSideNavigation("")` → false (`/^\/(?!\/)|^#/.test("")`)
- `isValidUrl("")` → false (`new URL("")` throws)

so it falls through to the final branch:

```js
/* Invalid URL */
return <MTypography component="span" variant="inherit">{label}</MTypography>;
```

That is a `<span>`, not an `<a>`. There is no `href`, nothing is clickable, and nothing enters the tab order — so the keyboard concern doesn't arise. I verified the built output too: no `href=""` appears anywhere in the generated page.

That said, the underlying point is fair: the cell is *relying* on a library fallback rather than saying what it means, and `url={doi ? … : ""}` reads like an oversight even though it behaves correctly.

**#3196 makes it explicit** and owns that decision — its description already calls this out ("the current implementation passes `url: ""` when the source dataset has no `doi`, producing a link that goes nowhere. Decide the no-DOI case explicitly — the recommendation is to render the citation as plain text with no link and no icon"). It has to branch on `doi` regardless, because the external-link icon it adds must not render without a DOI.

So leaving this PR as-is deliberately: no behaviour change is available to make, and the explicit branch belongs with the icon that motivates it. Not carrying the `url=""` pattern any further than it already goes.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
```diff
@@ -52,20 +47,13 @@ export function renderExplore(
...
-    label,
-    rel: REL_ATTRIBUTE.NO_OPENER_NO_REFERRER,
-    target: ANCHOR_TARGET.BLANK,
-    url: doi ? `${DOI_BASE_URL}${doi}` : "",
-  });
+export function renderFileName(
+  ctx: CellContext<TrackerSourceDataset, unknown>
+): JSX.Element {
+  return FileNameCell({ row: ctx.row.original });
 }
```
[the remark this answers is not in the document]
frano-m: Good catch — fixed in 2305d626, and you were right that I'd underweighted it.

`viewBuilder.ts` is now `viewBuilder.tsx` and all three renderers return real elements:

```tsx
return <TrackerDownloadCell {...props} />;
return <AnalysisPortalCell analysisPortals={analysisPortals} />;
return <FileNameCell row={ctx.row.original} />;
```

Also dropped the `import * as C` barrel namespace in favour of direct component imports while the file was open.

**The concern was more live than my `FileNameCell` alone.** `renderDownload` in the same file was calling `TrackerDownloadCell` as a function, and that component uses `useDialog()` — so its hooks were attaching to whichever component rendered the cell. It worked only because each renderer is invoked unconditionally and the cell count is stable. That's exactly the hazard you describe, and it already existed rather than being hypothetical for #3196. My earlier reasoning ("the file already does this, and my cell has no hooks") treated the existing instances as precedent when the hook-using one was the actual problem.

I went with the `.tsx` rename over `React.createElement`. Your point that `createElement` keeps the file `.ts` was the option I'd missed and it would have worked, but with three renderers converting, JSX reads better and the rename is import-transparent — every consumer imports `./viewBuilder` extensionless.

**Deliberately not fixed here:** the same pattern at 17 call sites in `viewModelBuilders/viewModelBuilders.ts`, including a second hook-using component (`CXGDownloadCell`, also `useDialog`). That file is 665 lines and drives the non-tracker atlas pages, the network tables and the home page, so the blast radius is well outside this issue. Filed as #3211 with the hook-ordering rationale.

Verified after the change: rendered content is unchanged — 7 file names, 7 DOI anchors, the CAP explore link, the download buttons, and no `href=""`. `tsc`, lint, prettier and `build-prod` all pass.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
```diff
@@ -0,0 +1,48 @@
...
+  // can never render empty.
+  const fileName = datasetAsset?.versionedFileName || baseFileName;
+  return (
+    <Stack spacing={2} useFlexGap>
+      <Typography variant={TYPOGRAPHY_PROPS.VARIANT.BODY_400}>
+        {fileName}
+      </Typography>
+      {publicationString && (
+        <Typography
+          color={TYPOGRAPHY_PROPS.COLOR.INK_LIGHT}
```
NoopDog, line 34: This `color` only takes effect on the no-DOI fallback. Whenever a DOI exists the citation renders as a MUI `Link`, which defaults to `color="primary"`, and findable-ui's theme doesn't override `MuiLink.root.color` — so the wrapper colour loses and the citation comes out primary blue rather than `ink.light`.

The description says the citation sits at `ink.light`, so one of the two should move: put the colour on the `Link` itself (`TypographyProps` / `sx`) if that's the intended design, or drop it from the wrapper if a blue link is what you actually want. Non-blocking either way, but right now the code and the description disagree.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
```diff
@@ -0,0 +1,48 @@
...
+          component="div"
+          variant={TYPOGRAPHY_PROPS.VARIANT.BODY_SMALL_400}
+        >
+          <Link
+            label={publicationString}
+            rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
+            target={ANCHOR_TARGET.BLANK}
+            url={doi ? `${DOI_BASE_URL}${doi}` : ""}
```
NoopDog, line 42: For a source study with no DOI, `url=""` sends findable-ui's `Link` into its invalid-URL branch, which spreads the remaining props onto a `<Typography component="span">` — emitting `<span rel="noopener noreferrer" target="_blank">`. Visually harmless, but invalid markup.

Carried over verbatim from the deleted `renderSourceStudy`, so not a regression from this PR. Passing `rel` / `target` only when `doi` is set (or rendering plain `Typography` in that case) would clear it up.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/accessor.ts
```diff
@@ -36,3 +36,14 @@ export function buildJournal(row: TrackerSourceDataset): string {
...
+
+/**
+ * Returns the file name shown in the pinned column, so the column sorts on the
+ * value it displays rather than on the unrevisioned base name. Mirrors the
+ * fallback in `FileNameCell` so no row ever sorts on an empty value.
+ * @param row - Tracker source dataset.
+ * @returns versioned file name, falling back to the base file name.
+ */
+export function buildVersionedFileNameValue(row: TrackerSourceDataset): string {
+  return row.datasetAsset?.versionedFileName || row.baseFileName;
```
NoopDog, line 48: This `versionedFileName || baseFileName` fallback is written twice — here and in `FileNameCell` (`fileNameCell.tsx:26`). They agree today, and the comments on both say so, but a later change to the displayed fallback would silently leave the column sorting on a value it no longer renders.

Having the cell call `buildVersionedFileNameValue(row)` would keep one source of truth.

NoopDog (approved):
Approving.

I read every hunk and traced the changed column config through findable-ui's table internals (`getColumnTrackSizing`, `getPinnedCellIndex`, `CollapsableCell`, `getFacetedUniqueValuesWithArrayValues`, `sortingFn`). `npx tsc --noEmit` and `eslint` are clean on the branch. No correctness bugs.

The things I specifically tried to break, and why they hold:

- **`sourceStudy: false` does not kill its filter.** Faceting and `getCanFilter` never consult visibility, the sidebar labels come from `META.categoryGroups` rather than column order, and the chips render filter *values* — so moving `SOURCE_STUDY` to the end of `COLUMNS` and dropping its `cell` / `meta.width` / `sortingFn` is inert. The deviation called out in the description is the right call: a hidden column's `cell` is never reached, so keeping `renderSourceStudy` would have left dead code, and the shape now matches the other filter-only columns.
- **Grid sizing** uses `table.getVisibleFlatColumns()`, so the removed width can't leave a dangling track; the remaining `fr` total is 6.4.
- **The new DOI link is a sibling** of the collapse `IconButton` in `PinnedCell`, not nested inside it — no nested-interactive or click conflict in the collapsed layout.
- **`initialState.sorting`** id `"fileName"` matches the new column id, and `accessorFn` + explicit `id` is valid.
- **The `baseFileName` fallback** is well-typed and can't be empty — it's required on `TrackerSourceDataset` and carries the extension.
- **The `.ts`→`.tsx` conversion is a real improvement**, not just cosmetic: `renderDownload` no longer calls the `useDialog`-bearing `TrackerDownloadCell` as a plain function.

Three low-severity notes inline — the first one contradicts the description's "citation at `ink.light`", so worth resolving one way or the other; the other two are tidy-ups, and neither blocks the merge.

One follow-up outside this diff: `viewModelBuilders/viewModelBuilders.ts:106` still calls `C.TrackerDownloadCell({...})` as a plain function after an early `return null` — the exact hook-in-a-function-call pattern this PR fixes here. That column is one `datasetAssets[0]` away from a hook-count mismatch, so it might be worth folding into #3205.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
frano-m: image

frano-m: You're right, and this was a real unmet criterion rather than a doc mismatch — thanks for catching it.

Confirmed from the emitted CSS:

```
wrapper .mui-nmk5ig : color: var(--palette-ink-light)     <- what I set
link    .mui-nw7vxf : color: var(--palette-primary-main)  <- what won
```

findable-ui's `MuiLink` override sets `cursor` and the `textDecoration*` properties but not `color`, so MUI's default `color="primary"` beat the inherited wrapper colour. The citation was rendering primary blue, and #3195's criterion ("visually secondary — `ink.light`, `body-small-400`") was only half met: the 13px came through via `font: inherit`, the colour did not.

Fixed by putting the colour on the link itself:

```tsx
TypographyProps={{ color: TYPOGRAPHY_PROPS.COLOR.INK_LIGHT }}
```

Now emits `color: var(--palette-ink-light)` on the anchor, verified in the built page and visually.

Worth being straight about how it got here: I asserted "citation at `ink.light`" in this PR's description on the strength of having written `color={INK_LIGHT}` on the wrapper, without checking what actually rendered. Every other criterion here I verified against the built HTML; this one I inferred from my own code, which is exactly the wrong way round.

The fix lands on the #3199 branch rather than here, since this PR is approved and green and #3199 sits on top of it — it reaches `main` immediately after.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
frano-m: Agreed it's invalid markup, and agreed it's inherited rather than introduced here. Two things came out of chasing it.

**One correction:** only `rel` leaks, not `target`. In `link.tsx` the signature destructures `target` but not `rel`:

```tsx
({ className, copyable = false, label, noWrap = false, onClick, target, TypographyProps, url, ...props })
```

so `target` never reaches the fallback, while `rel` stays in `...props` and gets spread onto the span. The output is `<span rel="noopener noreferrer">` with no `target`.

**Fixed in the consumer, and filed against the component.** #3196 (stacked on this PR) branches on `doi` and renders plain `Typography` when there is none, so no empty URL reaches `Link` at all — which also removes the `url=""` pattern Copilot flagged separately on this PR.

The component itself still has the sharp edge for any other caller, so that's DataBiosphere/findable-ui#1009: destructure `rel` the way `target` already is, plus the open question of whether the fallback branch should filter the spread to `Typography`-safe props rather than handling anchor-only props one at a time as they surface.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/accessor.ts
frano-m: Fixed — the cell now calls the accessor instead of repeating the expression:

```tsx
const fileName = buildVersionedFileNameValue(row);
```

So there is one definition of the displayed-and-sorted value, and the drift you describe (the column sorting on something it no longer renders) is no longer possible. `FileNameCell` no longer destructures `baseFileName` or `datasetAsset` at all, and the accessor's JSDoc now carries the reasoning about why the fallback exists rather than duplicating it in both files.

Landing on the #3199 branch with the other two notes, since this PR is approved and green.

PR #3209, opened 2026-09-09
feat: atlas source datasets table: add an external-link icon to the source study citation (#3196)

### LINKED ISSUE #3196: feat: atlas source datasets table: add an external-link icon to the source study citation

frano-m:
## Summary

Closes #3196. Part of #3193.

The source study citation opens `https://doi.org/{doi}` in a new tab but rendered as plain link text with no signal that it leaves the site. This appends `OpenInNewIcon` to the citation label, and makes the no-DOI case explicit.

## Changes

- `FileNameCell/fileNameCell.tsx` — the citation now branches on `doi`. With a DOI, the `label` is a fragment of the citation text followed by the icon, so **text and icon are a single link target**. Without one, the citation renders as bare `Typography` text: no link, no icon.
- `FileNameCell/fileNameCell.styles.ts` — **new**, `StyledOpenInNewIcon` with `margin-left: 4px` and `vertical-align: middle`.

## Two deviations from the issue, both because #3195 landed first

1. **No `.ts` → `.tsx` rename.** #3196 says to rename `viewBuilder.ts` and compose the label there. #3195 moved that rendering into `FileNameCell.tsx`, which is already JSX, so there is nothing to rename — the issue anticipates this ("If #3195 moves this rendering into a dedicated cell component, apply the change there instead").
2. **`OpenInNewIcon` imported from findable-ui directly, not via `C.OpenInNewIcon`.** The AC names `components/index.tsx`, but that barrel only re-exports the same findable-ui component, and importing the barrel from a file *inside* `components/` to build a `styled()` wrapper risks an import cycle. Same component either way; the cell already imports `Link` from findable-ui directly, so this is consistent within the file.

## The no-DOI fix, and a correction to Copilot's review of #3208

Copilot flagged the pre-existing `url={doi ? … : ""}` on #3208 as rendering "a clickable DOI link with an empty URL … problematic for keyboard users". **That was not accurate** — findable-ui's `Link` already degrades an invalid URL to a plain span. Traced through `link.js` with `url=""`:

- `isURLString("")` → true (`typeof value === "string"`)
- `isClientSideNavigation("")` → false (`/^\/(?!\/)|^#/`)
- `isValidUrl("")` → false (`new URL("")` throws)

falling through to `<MTypography component="span">{label}</MTypography>`. No anchor, no `href`, nothing focusable — and `href=""` appears nowhere in the generated output.

The underlying point still stood, though: the cell was *relying* on that fallback instead of saying what it meant. This PR removes the `url=""` pattern entirely, which #3196 asked for independently and which is required here anyway — the icon must not render without a DOI.

## Verification

Against the built page for Breast v1.0:

- **7 DOI anchors, all 7 containing an `<svg>`** — so every citation link carries the icon, and it is inside the anchor rather than beside it. Sample, tags collapsed: `<a>Gray (2022) Developmental Cell<svg><path></path></svg></a>`
- `target="_blank"` and `rel="noopener noreferrer"` on all 7
- **`href=""` count: 0**
- Icon colour needs no styling — `OpenInNewIcon` renders `fill="currentColor"`, so it inherits `ink.light` from the citation `Typography`, and defaults to `fontSize="xsmall"`
- Alignment and row height confirmed visually in the browser at desktop width
- `npm run lint` (0 errors), `npm run check-format`, `npx tsc --noEmit`, `npm run build-prod:data-portal` (64/64) all pass

## Not verifiable with live data

Breast v1.0 carries a DOI on all seven source datasets, so the no-DOI branch cannot be exercised against the real tracker. Its behaviour is asserted from the code (`doi` falsy → bare `Typography`, no `Link`, no icon) rather than observed — flagging that rather than implying it was tested.

image

### DIFF
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
@@ -2,9 +2,20 @@ import { OpenInNewIcon } from "@databiosphere/findable-ui/lib/components/common/
 import styled from "@emotion/styled";
 
 /**
+ * Holds the citation's final word and the external-link icon on one line. A
+ * browser may break before an atomic inline box, which would otherwise strand
+ * the icon on a line of its own when the citation wraps.
+ */
+export const StyledNoWrap = styled.span`
+  white-space: nowrap;
+`;
+
+/**
  * External-link affordance for the citation line. Inline so it flows with the
- * citation text and shares its colour via `currentColor`, and vertically
- * centred against the smaller secondary line without adding to its height.
+ * citation text and shares its colour via `currentColor`. The size is set with
+ * the `fontSize` prop at the call site rather than here: the theme's
+ * `MuiSvgIcon-fontSizeXsmall` variant wins over a `font-size` declaration in
+ * this block, so styling it here silently has no effect.
  */
 export const StyledOpenInNewIcon = styled(OpenInNewIcon)`
   margin-left: 4px;
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
@@ -1,0 +1,12 @@
+import { OpenInNewIcon } from "@databiosphere/findable-ui/lib/components/common/CustomIcon/components/OpenInNewIcon/openInNewIcon";
+import styled from "@emotion/styled";
+
+/**
+ * External-link affordance for the citation line. Inline so it flows with the
+ * citation text and shares its colour via `currentColor`, and vertically
+ * centred against the smaller secondary line without adding to its height.
+ */
+export const StyledOpenInNewIcon = styled(OpenInNewIcon)`
+  margin-left: 4px;
+  vertical-align: middle;
+`;
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
@@ -3,6 +3,7 @@ import {
   REL_ATTRIBUTE,
 } from "@databiosphere/findable-ui/lib/components/Links/common/entities";
 import { Link } from "@databiosphere/findable-ui/lib/components/Links/components/Link/link";
+import { SVG_ICON_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/svgIcon";
 import { TYPOGRAPHY_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/typography";
 import { Stack, Typography } from "@mui/material";
 import type { JSX } from "react";
@@ -45,7 +46,7 @@ export const FileNameCell = ({ row }: Props): JSX.Element => {
                   <StyledNoWrap>
                     {splitTrailingWord(publicationString).tail}
                     <StyledOpenInNewIcon
-                      fontSize="xxsmall"
+                      fontSize={SVG_ICON_PROPS.FONT_SIZE.XXSMALL}
                       titleAccess={EXTERNAL_LINK_TITLE}
                     />
                   </StyledNoWrap>
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
@@ -7,6 +7,7 @@ import { TYPOGRAPHY_PROPS } from "@databiosphere/findable-ui/lib/styles/common/m
 import { Stack, Typography } from "@mui/material";
 import type { JSX } from "react";
 import { DOI_BASE_URL } from "./constants";
+import { StyledOpenInNewIcon } from "./fileNameCell.styles";
 import type { Props } from "./types";
 
 /**
@@ -35,12 +36,23 @@ export const FileNameCell = ({ row }: Props): JSX.Element => {
           component="div"
           variant={TYPOGRAPHY_PROPS.VARIANT.BODY_SMALL_400}
         >
-          <Link
-            label={publicationString}
-            rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
-            target={ANCHOR_TARGET.BLANK}
-            url={doi ? `${DOI_BASE_URL}${doi}` : ""}
-          />
+          {doi ? (
+            // Text and icon are one label, so they are a single link target.
+            <Link
+              label={
+                <>
+                  {publicationString}
+                  <StyledOpenInNewIcon />
+                </>
+              }
+              rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
+              target={ANCHOR_TARGET.BLANK}
+              url={`${DOI_BASE_URL}${doi}`}
+            />
+          ) : (
+            // No DOI, so no link and no external-link affordance.
+            publicationString
+          )}
         </Typography>
       )}
     </Stack>
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
@@ -6,9 +6,11 @@ import { Link } from "@databiosphere/findable-ui/lib/components/Links/components
 import { TYPOGRAPHY_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/typography";
 import { Stack, Typography } from "@mui/material";
 import type { JSX } from "react";
-import { DOI_BASE_URL } from "./constants";
-import { StyledOpenInNewIcon } from "./fileNameCell.styles";
+import { buildVersionedFileNameValue } from "../../accessor";
+import { DOI_BASE_URL, EXTERNAL_LINK_TITLE } from "./constants";
+import { StyledNoWrap, StyledOpenInNewIcon } from "./fileNameCell.styles";
 import type { Props } from "./types";
+import { splitTrailingWord } from "./utils";
 
 /**
  * Pinned cell stacking the published file name over the source study citation.
@@ -19,12 +21,10 @@ import type { Props } from "./types";
  * @returns stacked file name and source study cell.
  */
 export const FileNameCell = ({ row }: Props): JSX.Element => {
-  const { baseFileName, datasetAsset, doi, publicationString } = row;
-  // `versionedFileName` is set for every tracker source dataset, but it is
-  // optional on the shared asset type - fall back to the unversioned base name
-  // so the pinned column, which identifies the row in the collapsed layout,
-  // can never render empty.
-  const fileName = datasetAsset?.versionedFileName || baseFileName;
+  const { doi, publicationString } = row;
+  // Shared with the column's `accessorFn`, so the pinned column always sorts on
+  // the value it displays.
+  const fileName = buildVersionedFileNameValue(row);
   return (
     <Stack spacing={2} useFlexGap>
       <Typography variant={TYPOGRAPHY_PROPS.VARIANT.BODY_400}>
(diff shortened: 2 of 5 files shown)


components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
```diff
@@ -0,0 +1,12 @@
+import { OpenInNewIcon } from "@databiosphere/findable-ui/lib/components/common/CustomIcon/components/OpenInNewIcon/openInNewIcon";
+import styled from "@emotion/styled";
+
+/**
+ * External-link affordance for the citation line. Inline so it flows with the
+ * citation text and shares its colour via `currentColor`, and vertically
+ * centred against the smaller secondary line without adding to its height.
```
NoopDog: The "without adding to its height" claim doesn't hold, and the icon ends up oversized.

`OpenInNewIcon` defaults to `fontSize="xsmall"`, which findable-ui's theme maps to **18px** (`lib/theme/common/components.js:1098`). The citation `Typography` is `BODY_SMALL_400` = **13px / 16px line-height** (`lib/theme/common/typography.js:38`). With `vertical-align: middle` the 18px box sits ~13.7px above and ~4.3px below the baseline, so the line box grows 16px → ~18px and the glyph renders at roughly 1.4× the text size.

Suggest an explicit smaller size — `fontSize="xxsmall"` (16px) or `font-size: inherit` — rather than taking the 18px default, and updating the comment either way.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+export const StyledOpenInNewIcon = styled(OpenInNewIcon)`
+  margin-left: 4px;
+  vertical-align: middle;
```
NoopDog, line 22: Worth checking in the browser: the link underline is likely drawn through the icon.

findable-ui sets `MuiLink.defaultProps.underline: "always"` together with `textDecorationSkipInk: "none"`, and this `Link` passes no `underline`/`TypographyProps` override, so the anchor is always underlined. Now that the SVG is inside the anchor, the underline runs under the 4px margin and across the icon — and because `vertical-align: middle` pushes the glyph ~4px below the baseline, the line cuts through the lower part of the box rather than passing beneath it.

If confirmed, either size/align the icon so it clears the underline, or set `underline="hover"` on this link.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
```diff
@@ -35,12 +36,23 @@ export const FileNameCell = ({ row }: Props): JSX.Element => {
...
-            label={publicationString}
-            rel={REL_ATTRIBUTE.NO_OPENER_NO_REFERRER}
-            target={ANCHOR_TARGET.BLANK}
-            url={doi ? `${DOI_BASE_URL}${doi}` : ""}
-          />
+          {doi ? (
+            // Text and icon are one label, so they are a single link target.
+            <Link
+              label={
+                <>
```
NoopDog, line 44: The icon can wrap onto a line by itself.

JSX strips the newline between `{publicationString}` and `<StyledOpenInNewIcon />`, so there's no text node between them, but browsers still allow a line break before an atomic inline-level box. This column is `min: "260px"` (`../table/columns.ts:86`), so a long citation — e.g. "Winkler (2024) Nature Reviews Molecular Cell Biology" — wraps, and the icon can end up alone on the final line, detached from the citation it annotates.

Wrapping the last word and the icon in a `white-space: nowrap` span avoids it.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
```diff
@@ -35,12 +36,23 @@ export const FileNameCell = ({ row }: Props): JSX.Element => {
...
+                  {publicationString}
+                  <StyledOpenInNewIcon />
```
NoopDog: The new cue is visual-only. MUI's `SvgIcon` sets `aria-hidden="true"` when no `titleAccess` is given, so a screen-reader user still gets no signal that this link opens in a new tab — the exact gap #3196 describes, closed for sighted users only.

If that's intentional (e.g. handled elsewhere), fine to leave. Otherwise `titleAccess` or visually-hidden text like "(opens in a new tab)" would close it for everyone.

NoopDog (approved):
Approving — the logic is correct and this is a real improvement over the old `url=""` fallback. The `doi ? … : …` branch is right (`doi` is `string | null`, and an empty string also lands on the no-link path), the icon is inside the anchor so text and icon are a single target, and `rel`/`target` are preserved. No correctness bugs.

Four low-severity rendering/a11y notes inline — none blocking, but 1 and 2 are worth a look in the browser before merge since they undercut the affordance being added.

One note on the PR description rather than the code: it says the icon "inherits `ink.light` from the citation `Typography`", but the icon is inside the anchor, so `currentColor` resolves to MUI `Link`'s own color (default `primary`), not `ink.light`.

Also confirming your read on the #3208 Copilot comment: it was inaccurate as traced — `isValidUrl("")` throws in `new URL("")`, so `link.js` falls through to the plain `MTypography` span.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
frano-m: image

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
frano-m: image

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
frano-m: image

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
frano-m: Fixed in 3d10c98c — and the route matters, because my first attempt silently didn't work.

Setting `font-size: 16px` in the styled block had **no effect**: the theme's `MuiSvgIcon-fontSizeXsmall` variant won, and the rendered icon stayed at 18px. I only caught it because I checked the emitted CSS afterwards rather than trusting the edit.

The size has to come from the prop:

```tsx
<StyledOpenInNewIcon fontSize="xxsmall" titleAccess={EXTERNAL_LINK_TITLE} />
```

`xxsmall` maps to 16px in the theme, and DevTools now reports the svg as `MuiSvgIcon-fontSizeXxsmall` at **16 × 16**, with the anchor 16.95px tall. The styled block keeps only `margin-left` and `vertical-align`, with a comment recording why the size is not set there.

Your numbers were right throughout, including the ~1.4× glyph. One correction to my own earlier framing rather than yours: I had described this as the icon exceeding "the line-height", when what actually governs is the line box formed from the parent's strut plus the atomic inline box — the anchor's own line-height doesn't come into it.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.styles.ts
frano-m: Checked in the browser as you suggested — **the underline does not cross the icon.**

Inspected the rendered anchor: the decoration runs under the citation text and stops; the icon sits clear of it. The anchor computes to 216.48 × 16.95 with the icon at 16 × 16 and `margin-left: 4px`, and `--Link-underlineColor` is applied to the text run rather than drawn across the svg box.

Your mechanism was sound — `underline: "always"` plus `textDecorationSkipInk: "none"` is exactly the combination that would do it, and with the icon at its 18px default sitting ~4px below the baseline it was the right thing to suspect. It just doesn't materialise here, and the size fix from your first note moves it further clear.

So no `underline="hover"` override, which I'd have been reluctant to add anyway — it would have dropped this one link out of the design system's always-underlined convention to solve something that isn't happening.

Caveat on method: my programmatic measurement attempt timed out (CDP `Runtime.evaluate` on a frozen renderer), so this rests on DOM inspection and zoomed screenshots rather than a scripted geometry check. Happy to redo it properly if you want the numbers.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
frano-m: Fixed in 3d10c98c, though worth saying it wasn't reproducing.

The citation's final word and the icon are now held together:

```tsx
{splitTrailingWord(publicationString).head}
<StyledNoWrap>
  {splitTrailingWord(publicationString).tail}
  <StyledOpenInNewIcon … />
</StyledNoWrap>
```

`splitTrailingWord` lives in the cell's `utils.ts`, and `StyledNoWrap` is a `white-space: nowrap` span. Rendered output confirms it: `…Developmental <span class="mui-uu6h57">Cell<svg …`.

**Observed behaviour before the fix:** the longest citation in Breast v1.0 (Twigger, "Twigger (2022) Nature Communications") already wraps at this column width, and the icon stayed attached to "Communications" on the second line rather than stranding. So the break you describe is possible but width-dependent, and the current data doesn't hit it.

Keeping the fix regardless — it converts "happens to be fine at these widths" into a guarantee, and the citation lengths here are set by whatever studies an atlas happens to cite.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
frano-m: Not intentional — you're right that it closed the gap for sighted users only. Fixed in 3d10c98c.

The icon now carries `titleAccess`, which makes `SvgIcon` emit a `<title>` and `role="img"` instead of `aria-hidden="true"`:

```ts
// constants.ts
export const EXTERNAL_LINK_TITLE = "(opens in a new tab)";
```

DevTools on the anchor now reports:

```
Name : Gray (2022) Developmental Cell (opens in a new tab)
Role : link
Keyboard-focusable : yes
```

So the cue is part of the link's accessible name rather than a separate element, which is what a screen reader user needs at the point of deciding whether to follow it. Verified in the built output too: 7 `<title>` elements and `role="img"` on all 7 citation icons, none of them `aria-hidden` any more.

Good catch — #3196 is explicitly about signalling that the link leaves the site, so shipping it as a visual-only cue would have half-delivered the issue.

frano-m: image

PR #3213, opened 2026-09-09
refactor: split the citation once in filenamecell (#3196)

frano-m:
## Summary

Small follow-up cleanup to #3196 (#3209), which has merged.

`FileNameCell` called `splitTrailingWord(publicationString)` twice in the JSX — once for `.head`, once for `.tail` — so the same string was split twice per row to render one citation.

```diff
+  const citation = publicationString
+    ? splitTrailingWord(publicationString)
+    : null;
 …
-      {publicationString && (
+      {citation && (
 …
-                  {splitTrailingWord(publicationString).head}
+                  {citation.head}
                   <StyledNoWrap>
-                    {splitTrailingWord(publicationString).tail}
+                    {citation.tail}
```

The split now also drives the citation's presence check, so `publicationString` is tested in one place rather than two.

## Why its own PR

The file is on `main` via #3209, and #3199 (Primary Data column) deliberately doesn't touch it — we'd just finished unpicking unrelated changes that had drifted onto that branch, so folding this in would have repeated the mistake for a five-line cleanup.

## Verification

Rendered output is unchanged: 7 DOI anchors, 7 `nowrap` spans each wrapping the final word plus its `<svg>`, and 7 `(opens in a new tab)` titles.

### DIFF
--- a/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
+++ b/components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/fileNameCell.tsx
@@ -26,12 +26,16 @@ export const FileNameCell = ({ row }: Props): JSX.Element => {
   // Shared with the column's `accessorFn`, so the pinned column always sorts on
   // the value it displays.
   const fileName = buildVersionedFileNameValue(row);
+  // Split once - the final word is held on one line with the icon that follows.
+  const citation = publicationString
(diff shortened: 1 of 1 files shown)


NoopDog (approved):
Behaviour-preserving and correctly scoped. `splitTrailingWord` always returns an object, so guarding on `citation` is equivalent to the previous `publicationString` guard. Verified `tsc --noEmit`, eslint and prettier clean on the branch; CI build passing.

Nit for a future pass, not worth holding this: guarding on `citation` is a slightly indirect stand-in for "we have a publication string" (the no-DOI branch still renders `publicationString` directly, and it stays `string | null` since narrowing is lost). Fine as-is.

Issue #3212, opened 2026-09-09
fix: analysis portal links give no 'opens in a new tab' cue and duplicate their icon's alt text


frano-m:
## Problem

Every analysis-portal link opens in a new tab, but nothing tells a screen-reader user that — and the portal's mark is announced twice.

`components/common/IconLink/iconLink.tsx` renders the image and the anchor as siblings, both labelled with the same string:

```tsx
<Stack className={className}>
  <StaticImage alt={label} height={height} src={icon} />
  <Link label={label} url={url} />
</Stack>
```

Rendered output for the tracker source datasets table:

```html
<img alt="Explorer" height="20" src="/hca-bio-networks/network/atlas/icons/hca.webp"/>
<a href="https://explore.data.humancellatlas.org/projects/…" rel="noopener noreferrer" target="_blank">Explorer</a>

<img alt="CAP" height="20" src="/hca-bio-networks/network/atlas/icons/cap.svg"/>
<a href="https://celltype.info/project/1041/dataset/3421" rel="noopener noreferrer" target="_blank">CAP</a>
```

Two distinct issues:

1. **No new-tab cue.** findable-ui's `Link` defaults external URLs to `target="_blank"`, so every one of these leaves the site with no warning for anyone not seeing the layout.
2. **Redundant alt text.** The `alt` duplicates the adjacent anchor text, so the mark is announced as an image *and* as the link name ("Explorer image, Explorer link"). The mark is decorative here — the anchor text already names the portal — so `alt=""` is the correct treatment.

## Affected surfaces

`IconLink` is consumed by `AnalysisPortalCell`, so this covers every portal link on the site:

- **Explore** column on the tracker source datasets table (CAP)
- **Primary Data (fastqs)** column on the same table (HCA Data Explorer)
- The **integrated objects** table on the atlas overview page (CAP, CZ CELLxGENE, UCSC Cell Browser)

## Suggested fix

In `IconLink`:

- `alt=""` on the `StaticImage`, so the decorative mark stops competing with the link text.
- Fold the new-tab cue into the link's accessible name, the way the source study citation now does — visually-hidden text, or a `title`/`aria-label` that appends it.

The citation link on the source datasets table already handles this via `titleAccess` on its `OpenInNewIcon`, producing an accessible name of "Gray (2022) Developmental Cell (opens in a new tab)". Worth landing something consistent, since right now the same table row treats its DOI link and its portal links differently.

## Acceptance criteria

[...]

PR #3214, opened 2026-09-09
feat: atlas source datasets table: add a primary data column linking to the hca explorer (#3199)

### LINKED ISSUE #3199: feat: atlas source datasets table: add a Primary Data column linking to the HCA Data Explorer

frano-m:
## Summary

Closes #3199. Part of #3193 — **the last sub-issue in the epic.**

Every source dataset derives from a source study whose primary data is often already in the HCA Data Explorer. This adds a **Primary Data (fastqs)** column linking each row there.

The `hcaProjectId` comes from the source-study join #3197 introduced, so no new fetch or derivation.

## ⚠️ Deviation: text-only link, no portal mark

#3199 asks for a *branded portal link* — "the HCA blue Voronoi-sphere mark with the label **Explorer**", "presented as a branded portal link rather than a bare icon" — with an acceptance criterion requiring a standalone HCA mark asset be added. Epic #3193's presentation conventions settled the same thing for named portals.

**That is not what this PR does.** At @NoopDog's request the icon is dropped and the cell renders **"HCA Explorer"** as plain text. Consequently the following are *not* in this PR, and the earlier revisions that added them have been reverted:

- `public/hca-bio-networks/network/atlas/icons/hca.webp` — the icon asset
- `HCA_DATA_EXPLORER` in `constants/analysisPortals.ts`
- `ANALYSIS_PORTAL.HCA_DATA_EXPLORER`
- the `AnalysisPortalCell` / `IconLink` treatment for this column (still used for CAP in the Explore column)

So #3199's mark-and-label criteria are deliberately unmet. The issue should be updated to match, or this recorded as a superseding decision.

## Changes

- `utils/network.ts` — `buildHCADataExplorerProjectUrl(browserURL, projectId)`, owning the `/projects/{id}` path.
- `viewBuilder.tsx` — `buildRenderPrimaryData(browserUrl)` returns the cell renderer, so the URL travels in a typed closure. Renders a plain `Link`; findable-ui's `Link` sends external URLs to a new tab with `rel="noopener noreferrer"` automatically.
- `columns.ts` — `COLUMNS` becomes **`getColumns(browserUrl)`**, matching the `getProjectsTableColumns` factory pattern one directory over. The Primary Data column sits beside Explore and Download.
- `utils.ts` — the column is hidden when no dataset in the atlas has an HCA project, or the environment has no browser URL.
- `hook.ts` — reads config through `hooks/useSiteConfig` like every other consumer in the repo.
- `constants.ts` — the `HCA Explorer` label.
- Plus the **File Name / Study** header rename, now that the pinned column carries both.

## From review

- **The table-meta contract is gone, not patched.** The browser URL was originally threaded through `table.options.meta` and read back via `as SourceDatasetsTableMeta`. The producer side was type-safe (dropping the key errors with `TS1360`, renaming it with `TS2561`), but the **consumer** side was an unchecked cast — reusing the renderer in a table whose meta lacked the key would have compiled and rendered `null` for every row, a visible header over empty cells with no error. The column factory removes the interface, the meta entry and the cast entirely.
- **Column width.** `width: "auto"` sized the column to its widest content, which was the header rather than the cells. Now `{ max: "1fr", min: "120px" }`, so the header wraps instead of driving the width. Confirmed in the generated grid: `minmax(120px, 1fr)` for the Primary Data track.
- **Config access** routed through `hooks/useSiteConfig` rather than findable-ui's `useConfig` directly.
- **Columns and visibility are memoised.** The column factory above returns a new array each render, and TanStack keys its `getAllColumns` memo on `options.columns` by identity — so without memoisation every render rebuilt the column instances and discarded their faceting caches, recomputing facets for all seven filter columns on each filter toggle, sort and breakpoint change. `useMemo` on both `getColumns(browserURL)` and `getColumnVisibility(data, browserURL)`.

## Verification

Against the built export for Breast v1.0:

- Header `Primary Data (fastqs)` between Cell Count and Explore; pinned column reads `File Name / Study`.
- **6 links across 7 rows** — Reed's source study has `hcaProjectId: null` and renders an empty cell, the AC's specific null case.
- 6 × `HCA Explorer`, and **zero** `hca.webp` references anywhere; the asset is absent from `out/`.
- Kumar resolves to `/projects/1ffa2223-28a6-4133-a5a4-badd00faf4bc`, the AC's named check.
- Anchors carry `target="_blank"` and `rel="noopener noreferrer"`.
- URLs come from the environment's configured `browserURL`, never hardcoded.
- The Explore column's CAP link and icon are untouched.
- **Narrow viewport confirmed** — the long `Primary Data (fastqs)` header wraps rather than forcing the column wider than its contents need, closing #3199's last acceptance criterion. Checked visually; the `minmax(120px, 1fr)` track is what makes it possible, since `width: "auto"` previously let the header set the width.

## Related, not fixed here

#3212 covers the missing "opens in a new tab" cue on portal links. With the icon removed, this link no longer has the duplicate-`alt` half of that problem — but it also no longer has an icon to carry `titleAccess`, so closing the cue for this column will need visually-hidden text. Worth noting a single row still treats its DOI citation link (which does announce the new tab, via #3196) differently from this one.

image


components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
```diff
@@ -58,3 +60,23 @@ export function renderFileName(
...
+): JSX.Element | null {
+  const { hcaProjectId } = ctx.row.original;
+  if (!hcaProjectId) return null;
+  const meta = ctx.table.options.meta as SourceDatasetsTableMeta | undefined;
+  if (!meta?.browserUrl) return null;
+  const analysisPortal = buildHCADataExplorerAnalysisPortal(
+    meta.browserUrl,
+    hcaProjectId
+  );
+  return <AnalysisPortalCell analysisPortals={[analysisPortal]} />;
```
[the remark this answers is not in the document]
frano-m: Correct on both counts, and already tracked — this is the gap declared in the PR description and filed as #3212.

Your diagnosis matches what I found in the rendered output:

```html
<img alt="Explorer" height="20" src="/hca-bio-networks/network/atlas/icons/hca.webp"/>
<a href="https://explore.data.humancellatlas.org/projects/…" rel="noopener noreferrer" target="_blank">Explorer</a>
```

The `alt` duplicates the anchor text (so the mark is announced as an image *and* as the link name), and `target="_blank"` carries no cue. #3212 proposes exactly what you suggest: `alt=""` on the `StaticImage` so the mark is decorative, plus folding the new-tab cue into the link's accessible name.

**Why it isn't fixed here.** The problem is in `components/common/IconLink/iconLink.tsx`, which `AnalysisPortalCell` uses for CAP, CZ CELLxGENE and UCSC Cell Browser as well — including on the atlas overview page's integrated-objects table. So the fix has a blast radius well beyond this column and wants verifying on a surface this PR doesn't touch. This PR adds a further instance of a pre-existing gap rather than introducing it.

That's a deliberate line rather than an oversight: this branch had unrelated changes drift onto it earlier in review and we unpicked them, so reaching into a shared component mid-ticket is the specific thing I'm avoiding.

Worth flagging what it leaves in the meantime: a single table row now treats its DOI citation link, which *does* announce the new tab via #3196's `titleAccess`, differently from its two portal links, which don't. That inconsistency is in the PR description and is the strongest argument for #3212 being picked up soon.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
```diff
@@ -58,3 +60,23 @@ export function renderFileName(
...
+ * @param ctx - Cell context.
+ * @returns AnalysisPortalCell, or null when the source study has no HCA
+ * project or the environment has no configured browser URL.
+ */
+export function renderPrimaryData(
+  ctx: CellContext<TrackerSourceDataset, unknown>
+): JSX.Element | null {
+  const { hcaProjectId } = ctx.row.original;
+  if (!hcaProjectId) return null;
+  const meta = ctx.table.options.meta as SourceDatasetsTableMeta | undefined;
```
NoopDog: The meta contract is only half-closed here, so the "From review" section overstates the fix.

`satisfies` on the producer catches a **renamed** key — which is the demonstration the description gives — but not a **dropped** one. Delete `browserUrl` from the spread in `hook.ts` and this cast still compiles, while every cell renders `null`: a visible header over empty cells, no error anywhere. Same for reusing `renderPrimaryData` in another table whose meta lacks the key. That's the exact failure mode the change set out to eliminate.

Closing it properly needs a `declare module "@tanstack/react-table"` augmentation of `TableMeta`, or the approach in my comment on `hook.ts`.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
```diff
@@ -17,6 +18,8 @@ import { getColumnVisibility } from "./utils";
 export const useTable = (
   data: TrackerSourceDataset[]
 ): Table<TrackerSourceDataset> => {
+  const { config } = useConfig();
```
NoopDog: The sibling table already solves this without touching meta at all — `getProjectsTableColumns(browserURL)` at `viewModelBuilders/viewModelBuilders.ts:459` is a column-def factory that closes over the URL in the cell renderer.

A `getColumns(browserUrl)` here would carry it in a typed closure and delete the `SourceDatasetsTableMeta` interface, the meta spread below, the cast in `viewBuilder.tsx`, and the `!meta?.browserUrl` branch (already excluded by `getColumnVisibility`). Smaller than what's here now, fully type-checked, and matches the pattern one directory over. This is the change I'd suggest.

Separately: this calls findable-ui's `useConfig` directly, while every other config consumer in the repo goes through `hooks/useSiteConfig` — `components/Home/components/SectionHero/sectionHero.tsx:22`, `components/Layout/components/Content/components/Link/link.tsx:17`, and the sibling `Datasets/components/MainColumn/mainColumn.tsx:13`. It works, since `browserURL` is on the base `SiteConfig`, but it's the only place that bypasses the local hook.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/columns.ts
```diff
@@ -103,6 +104,16 @@ const JOURNAL = {
...
 } as ColumnDef<TrackerSourceDataset>;
 
+const PRIMARY_DATA = {
+  accessorKey: "hcaProjectId",
+  cell: renderPrimaryData,
+  enableColumnFilter: false,
+  enableSorting: false,
+  header: "Primary Data (fastqs)",
+  id: "primaryData",
+  meta: { width: "auto" },
```
NoopDog: One acceptance criterion on #3199 looks unverified: "The header text is long relative to its cell contents — check it does not force the column wider than the mark-plus-label needs, and wraps or truncates acceptably at narrow viewports."

With `width: "auto"`, `Primary Data (fastqs)` is the widest thing in the column, so the header drives its width. Every other AC is covered in the description, and the screenshot is a single wide viewport — could you confirm this one at a narrow width?

NoopDog (commented):
Reviewed the full diff and checked the surrounding code: the tracker join that produces `hcaProjectId`, the dev/prod `browserURL` config, `IconLink`/`AnalysisPortalCell`, and the single consumer of `COLUMNS`/`getColumnVisibility`/`META`. `tsc --noEmit` clean on a worktree of the head, Prettier clean, CI green.

**No correctness bugs, and nothing here blocks merge.** The null path is handled twice over, and the `browserUrl` guard is better justified than it first looks — an empty `browserURL` would produce a *relative* `/projects/{id}` href that navigates inside the portal rather than to the Explorer, so that guard prevents a real (if unlikely) misconfiguration failure.

I verified the asset claims independently: 128×128 lossy WebP, 7,334 bytes, `ALPH` chunk present (so no white box behind the sphere), and the 1500×1500 RGBA source really does already exist at `public/hca-bio-networks/about/mosaic-ball.png`. WebP over the issue's "preferably SVG" is a documented, quantified deviation — fine by me.

Three comments inline. The second is the one I'd actually act on; it subsumes the first.

Two notes needing no action:

- The `File Name` → `File Name / Study` rename belongs to #3195/#3208 rather than #3199. Disclosed, one line, and it does make the header honest — just noting the bundling.
- On the #3212 gap: the link *does* open in a new tab (findable-ui's `Link` defaults external URLs to `target=_blank`). What's missing is the *cue* to screen-reader users, plus `alt={label}` duplicating the adjacent anchor text. Accurately described and correctly scoped out.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
frano-m: Done — this is a better change than what I had, and it deletes more than it adds. Implemented in 08a85b3a.

`COLUMNS` is now `getColumns(browserUrl)`, and `renderPrimaryData` became `buildRenderPrimaryData(browserUrl)` returning the cell renderer, so the URL travels in a typed closure. That removed, as you predicted:

- the `SourceDatasetsTableMeta` interface
- the meta spread in `hook.ts` (back to plain `meta: META`)
- the `as SourceDatasetsTableMeta | undefined` cast in `viewBuilder.tsx`
- the `!meta?.browserUrl` branch, which `getColumnVisibility` already covered

`grep` for `SourceDatasetsTableMeta` and `options.meta` in that directory now returns nothing.

Also switched to `hooks/useSiteConfig` — you're right that this was the only consumer bypassing it. Its `browserURL` is typed on `SiteConfig` and it throws if the config is missing `portalURL`, so it's a stronger contract than `useConfig` as well as a consistent one.

Separately, and unprompted, the icon is gone from this column entirely: it now renders "HCA Explorer" as plain text per your no-image request, so the portal constant, its `ANALYSIS_PORTAL` enum member, and the `hca.webp` asset are all removed too. That deviates from #3199's stated presentation, which I've noted in the description.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
frano-m: Resolved, but by removing the contract rather than closing it — and one detail of the diagnosis doesn't hold.

**The producer half was already covered.** I tested dropping `browserUrl` from the spread, as described:

```
error TS1360: Type '{ categoryGroups?: CategoryGroup[]; filterSort?: FILTER_SORT; }'
does not satisfy the expected type 'SourceDatasetsTableMeta'.
```

`satisfies` catches a dropped required key as well as a renamed one, since the remaining object no longer satisfies the type. So that specific failure mode wasn't open.

**Your underlying point was right, though**, and it was about the other side: `ctx.table.options.meta as SourceDatasetsTableMeta | undefined` is an unchecked cast, so nothing linked the consumer to any producer. Reusing `renderPrimaryData` in a table whose meta lacks the key would have compiled and rendered `null` for every row — exactly the silent-empty-column outcome the change was meant to prevent, just reached from the consumer rather than the producer.

**Rather than adding a `declare module` augmentation, the meta is gone.** Following your `hook.ts` suggestion, the browser URL now travels in a typed closure via `getColumns(browserUrl)`, so there is no meta entry, no interface and no cast to get wrong. The `TableMeta` augmentation would have worked, but it widens the type for every table in the app to carry a field only this one needs.

The description's "From review" section overstated the fix, as you say — it's been rewritten to describe the closure instead.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/columns.ts
frano-m: You're right that this one was unverified — I'd covered the other criteria against the built output and left this one on the strength of a single wide screenshot.

Two things have changed since, both bearing on it:

**1. The cell contents shrank.** The icon is gone at your request, so the cell is now the text "HCA Explorer" rather than a mark plus label — which makes the header *more* dominant, not less, and the criterion more pointed.

**2. The column no longer uses `width: "auto"`.** It now carries a min/max, so the header wraps instead of setting the column width:

```ts
meta: { width: { max: "1fr", min: "120px" } }
```

Confirmed in the generated grid — the Primary Data track is the sixth:

```
minmax(260px, 2.4fr)  File Name / Study
minmax(100px, 1fr)    Assay
minmax(100px, 1fr)    Tissue
minmax(100px, 1fr)    Disease
minmax(100px, 1fr)    Cell Count
minmax(120px, 1fr)    Primary Data (fastqs)   <- was `auto`
auto                  Explore
auto                  Download
```

`auto` was the actual cause of the concern: it sizes to content, and the header was the widest content. With a 120px floor and a `1fr` share the header has to wrap, and 120px comfortably fits "HCA Explorer" on one line so the cells don't wrap.

**What I still haven't done is look at it.** The grid math says the header wraps rather than forcing width, but "wraps or truncates acceptably" is a visual judgement, and I'd rather not claim it from CSS values alone the way I did last time. Requesting a narrow-viewport check before this merges — flag it if the wrapped header reads badly and I'll adjust the floor.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
```diff
@@ -17,8 +18,10 @@ import { getColumnVisibility } from "./utils";
 export const useTable = (
   data: TrackerSourceDataset[]
 ): Table<TrackerSourceDataset> => {
+  const { browserURL } = useSiteConfig();
+
   return useReactTable<TrackerSourceDataset>({
-    columns: COLUMNS,
+    columns: getColumns(browserURL),
```
NoopDog: `getColumns(browserURL)` returns a brand-new array on every render — new inline column def, new `buildRenderPrimaryData` closure — where the old `COLUMNS` constant was reference-stable.

TanStack's `getAllColumns` memo is keyed on `options.columns` by reference, so every render now discards and rebuilds all column instances, taking their per-column memo caches with them. In practice that means `getFacetedRowModel`/`getFacetedUniqueValues` get recomputed for all seven filter columns on each render — every filter toggle, sort click and breakpoint change.

```ts
const columns = useMemo(() => getColumns(browserURL), [browserURL]);
```

Same applies to `getColumnVisibility(data, browserURL)` on line 38 — that one was already unstable before this PR, so fixing it here is optional, but it's the same one-line change.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/utils.ts
```diff
@@ -4,18 +4,26 @@ import type { TrackerSourceDataset } from "../../../../../../../../../../@types/
...
+  data: TrackerSourceDataset[],
+  browserUrl: string
 ): VisibilityState {
   return {
     explore: data.some(({ capUrl }) => Boolean(capUrl)),
     integratedObject: false,
     journal: false,
+    primaryData:
+      Boolean(browserUrl) &&
+      data.some(({ hcaProjectId }) => Boolean(hcaProjectId)),
```
NoopDog, line 28: Two small things about the `Boolean(browserUrl) &&` half of this guard:

1. It can't fire — `browserURL` is a required `string` on `SiteConfig` and is set in both the dev and prod configs, so it's dead.
2. The doc comment above says this matches "the conditions `renderPrimaryData` renders under", but `renderPrimaryData` doesn't check `browserUrl` at all — it only early-returns on a falsy `hcaProjectId`.

So if the config ever *did* yield an empty `browserURL`, the two halves would disagree: this guard hides the column, while the renderer would happily emit relative `/projects/<id>` hrefs, which `Link` treats as client-side navigation to a route that doesn't exist on this site.

Either drop the `Boolean(browserUrl)` term (and the matching clause in the comment), or add the same check to `renderPrimaryData` so the claim holds.

NoopDog (commented):
Hi @frano-m — reviewed this one hunk by hunk, plus the surrounding table plumbing (`hook.ts`, `meta.ts`, the common `Table`, findable-ui's `Link`, `getColumnTrackSizing`, the tracker join in `utils/trackerAtlasPages.ts`, and both site configs).

The core logic checks out and I found no correctness bug. Specifically verified: `hcaProjectId` really is on `TrackerSourceDataset` (joined in `getTrackerContentStaticProps`, `null` on a join miss); the null case returns `null` like the sibling `renderExplore`/`renderDownload` renderers; `browserURL` is a required `string` in `SiteConfig` and neither the dev nor the prod value has a trailing slash; findable-ui's `Link` does route an absolute URL to `target="_blank"` + `rel="noopener noreferrer"`; the `primaryData` visibility key matches the column `id`; no other consumer of the removed `COLUMNS` export exists; and the added `1fr` track keeps the grid's total fractional size above 1, so `getColumnTrackSizing` won't rewrite the first column.

Three low-severity things — two inline, and one that falls outside the changed files:

**`viewModelBuilders/viewModelBuilders.ts:504` — the `/projects/{id}` path now lives in two places.** `getProjectTitleUrl` still hardcodes ``` `${browserURL}/projects/${processEntityValue(...)}` ```. The description says `buildHCADataExplorerProjectUrl` owns that path, but this pre-existing caller wasn't switched over, so a future Explorer route change would silently fix only one of the two tables. Cheap to route it through the new helper while it's fresh.

The deliberate deviation from #3199 (text-only "HCA Explorer", no branded mark) reads as a decision to record on the issue rather than anything to change here.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/hook.ts
frano-m: Fixed in e38c18ea — and this was a regression I introduced by taking your earlier suggestion without thinking through what it cost.

Swapping the reference-stable `COLUMNS` constant for a `getColumns(browserURL)` call did exactly what you describe: a new array, a new inline column def and a new `buildRenderPrimaryData` closure on every render, so `getAllColumns` invalidated and every column's faceting cache went with it. Seven filter columns recomputing on each filter toggle, sort click and breakpoint change — measurably worse than the untyped meta it replaced.

Both are now memoised:

```ts
const columns = useMemo(() => getColumns(browserURL), [browserURL]);
const columnVisibility = useMemo(
  () => getColumnVisibility(data, browserURL),
  [data, browserURL]
);
```

I took the optional half too. You're right that `getColumnVisibility` was already unstable before this PR, but it's the same one-line change and leaving it would have meant `state.columnVisibility` re-applying on every render for no reason.

Rendered output is unchanged: same eight headers, 6 `HCA Explorer` links across 7 rows, and the grid still `minmax(120px, 1fr)` for the Primary Data track.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/utils.ts
frano-m: The comment was wrong and is fixed. I've kept the guard, though — and I think the failure mode you describe can't actually occur.

**You're right on both counts about the comment.** It claimed to match "the conditions `renderPrimaryData` renders under", which was true when the renderer early-returned on `!meta?.browserUrl` — but I removed that branch in the same change that moved to the closure, and left the comment describing code that no longer exists. And `Boolean(browserUrl)` is indeed unreachable, since `SiteConfig.browserURL` is a required string set in both environments.

**Where I'd push back is the disagreement scenario.** If `browserURL` were ever empty, this guard sets `primaryData: false` — and a hidden column's cells are never rendered, so `renderPrimaryData` doesn't run and no relative `/projects/{id}` href is produced. The two halves can't diverge, because the visibility guard is what prevents the renderer from executing at all.

That's demonstrable in this very table: `JOURNAL`, `REFERENCE_AUTHOR` and `SOURCE_STUDY` are hidden and have **no `cell` property whatsoever**. If hidden columns rendered cells, those three would blow up on every row.

So rather than dropping the term or duplicating the check into the renderer, I've kept the guard as the single place the condition lives and rewritten the comment to say what it actually is:

```
The `browserUrl` term is defensive rather than reachable - `SiteConfig.browserURL`
is a required string set in both environments - but hiding the column is what stops
`renderPrimaryData` from ever building a relative `/projects/{id}` href, since a
hidden column's cells are not rendered.
```

Happy to drop the term entirely if you'd rather have no unreachable code; my reasoning for keeping it is that `browserURL: ""` is expressible in a config today with nothing to catch it, and a hidden column degrades better than a broken client-side link.

NoopDog (approved):

PR #3216, opened 2026-09-09
refactor: render viewmodelbuilders cells as elements (#3211)

### LINKED ISSUE #3211: refactor: render cells as elements in viewModelBuilders rather than calling components as functions
## Problem

`viewModelBuilders/viewModelBuilders.ts` builds table cells by invoking components as plain functions rather than creating elements:

```ts
return C.TrackerDownloadCell(props);
C.CXGDownloadCell({ datasetAssets: …, title: … });
C.NTagCell({ label: "assays", values: row.original.assay });
// …17 call sites in total
```

Calling a component this way inlines its body into whichever component is rendering the cell. There is no component instance and no reconciliation boundary, so hooks, context, `memo` and refs do not behave as they would for a real element.

**Two of those components use hooks.** `TrackerDownloadCell` and `CXGDownloadCell` both call `useDialog()`, so their hooks currently attach to the enclosing renderer's hook list. It works today only because each cell renderer is invoked unconditionally and the cell count is stable — but it breaks the moment a call becomes conditional, or a hook is added behind an early return. The remaining 15 components (`NTagCell`, `PinnedNTagCell`, `AnalysisPortalCell`, `BasicCell`, `BioNetworkCell`, `ButtonOutline`, `Link`) use no hooks, so those sites are a consistency and correctness-of-idiom concern rather than a live risk.

## Approach

Rename `viewModelBuilders/viewModelBuilders.ts` to `.tsx` and return JSX from every cell renderer:

```tsx
return <TrackerDownloadCell {...props} />;
```

Every consumer imports it extensionless (`from ".../viewModelBuilders/viewModelBuilders"`), so the rename needs no import changes anywhere.

While in there, drop the `import * as C` barrel namespace in favour of direct component imports, which is what the equivalent file for the tracker source datasets table now does.

## Scope note

The same pattern in the tracker source datasets table's own view builder is already fixed — that file is `.tsx`, returns real elements, and imports components directly. This issue is the remaining occurrence, and it is the larger one: a 665-line file that also drives the non-tracker atlas pages, the network tables and the home page, so the blast radius is wider than the tracker work and it deserves its own change rather than riding along.

## Acceptance criteria

- [ ] `viewModelBuilders.tsx` returns elements from every cell renderer; no component is called as a function.
- [ ] `TrackerDownloadCell` and `CXGDownloadCell` in particular render as elements, so their `useDialog` hooks sit inside their own component instances.
- [ ] No `import * as C` remains in the file.
- [ ] No import statements elsewhere needed changing (the rename is extensionless-transparent).
- [ ] The non-tracker atlas source studies table, the network atlases tables and the home page render unchanged — compare the generated `out/` content, not file hashes (the build is not byte-deterministic between runs).
- [ ] Download dialogs still open and show the right file name and size on both a tracker atlas and a CXG-sourced one.

## Notes

Raised in review of the atlas source datasets work. No user-visible bug exists today; the value is removing the latent hook-ordering hazard around the two dialog cells and making the file idiomatic.

Worth doing before anything adds a hook to one of the other 15 components, since nothing would flag it.

frano-m:
## Summary

Closes #3211.

`viewModelBuilders.ts` built table cells by invoking components as plain functions — `C.NTagCell({ … })` rather than `<NTagCell … />` — across 17 call sites. That inlines the component body into whichever component renders the cell, so there is no component instance and no reconciliation boundary: hooks, context, `memo` and refs don't behave as they would for a real element.

**Two of those components use hooks.** `TrackerDownloadCell` and `CXGDownloadCell` both call `useDialog()`, so their hooks were attaching to the enclosing renderer's hook list. It held only because each cell renderer runs unconditionally and the cell count is stable — the same latent hazard that was fixed for the tracker source datasets table in #3195.

## Changes

- **Renamed** `viewModelBuilders/viewModelBuilders.ts` → `.tsx`. Every consumer imports it extensionless, so no import statement anywhere needed changing.
- **All 17 cell renderers now return elements.** `TrackerDownloadCell` and `CXGDownloadCell` (the two with hooks) plus `NTagCell` ×5, `PinnedNTagCell` ×2, `BasicCell` ×3, `Link` ×2, `AnalysisPortalCell`, `BioNetworkCell`, and the `ButtonOutline` in `buildCookieBanner`.
- **Dropped the `import * as C` barrel** for direct imports — four from findable-ui (`CookieBanner`, `Link`, `BasicCell`, `NTagCell`) and six local (`ButtonOutline`, `PinnedNTagCell`, `BioNetworkCell`, `CXGDownloadCell`, `TrackerDownloadCell`, `AnalysisPortalCell`).

## Verification

This file drives more than the atlas pages, so I compared the **whole** static export before and after rather than spot-checking.

Captured a structural fingerprint of all 63 exported pages on both `main` and this branch — download-button count, chip count, the full set of anchor `href`s and the full set of ``s per page:

```
pages before/after: 63 / 63
pages with structural differences: 0
```

**Identical across every page** — every link target, every image, every download button and every N-tag chip. Surfaces exercised: the tracker atlas overview (integrated objects, CAP portal links), the non-tracker atlas `/datasets` page (Azul projects table, 49 chips), the network atlases tables, the bio-networks list, and the home page.

Note I compared *content* rather than file hashes deliberately: **this build is not byte-deterministic** — identical code produces different HTML hashes between runs, so hash comparison reports false differences.

## No conflicts

Checked all 41 open PRs before starting: none touches `viewModelBuilders`, which matters because the `.ts` → `.tsx` rename would conflict with any concurrent edit to the file. #3213 and #3214 touch the source datasets table only.

## Not a repo-wide sweep - 6 sites remain, filed as #3217

This PR fixes `viewModelBuilders` only. A full audit afterwards found **6 more component-as-function calls**, all in `site-config/data-portal/dev/config.ts`:

| Line | Call |
| --- | --- |
| 87 | `Branding: C.Logo({…})` |
| 117 | `logo: C.Logo({…})` |
| 184, 188, 192, 196 | `icon: C.GitHubIcon / FacebookIcon / XIcon / LinkedInIcon({ fontSize: "small" })` |

**They are benign, for two reasons that don't apply to this PR's sites.** None of those five components uses a hook, so there is no hook-ordering hazard — unlike `TrackerDownloadCell` and `CXGDownloadCell` here, which both call `useDialog()`. And they run at **module scope** rather than during a render, so they are a different shape of problem: elements built eagerly at import time and stored in config, not components inlined into a parent's render.

Filed as #3217 because "config constructs elements eagerly" is a pattern worth a deliberate decision rather than being quietly converted — it also constrains those components to stay hook-free, which nothing currently enforces.

Four other modules still import the `C` barrel and want the same audit: `mdx-components.tsx`, `docs/common/constants.ts`, `site-config/data-portal/dev/socialMedia.ts`, `site-config/data-portal/dev/layout/floating.ts`. A sweep for non-barrel direct calls (`PascalCase({ … })` in `.ts`, `return PascalCase({ … })` in `.tsx`) came back clean.

## One fix worth noting

The `CXGDownloadCell` conversion needed parentheses added around the JSX in a concise arrow body — the original `=> C.CXGDownloadCell({…})` had its closing paren belonging to the call, so converting the call without adding the wrapper left an unbalanced paren. `tsc` caught it immediately.

### DIFF
--- a/viewModelBuilders/viewModelBuilders.tsx
+++ b/viewModelBuilders/viewModelBuilders.tsx
@@ -1,10 +1,16 @@
 import { LABEL } from "@databiosphere/findable-ui/lib/apis/azul/common/entities";
+import { CookieBanner } from "@databiosphere/findable-ui/lib/components/common/Banner/components/CookieBanner/cookieBanner";
 import { KeyValues } from "@databiosphere/findable-ui/lib/components/common/KeyValuePairs/keyValuePairs";
 import {
   ANCHOR_TARGET,
   REL_ATTRIBUTE,
 } from "@databiosphere/findable-ui/lib/components/Links/common/entities";
-import { MetadataValue } from "@databiosphere/findable-ui/lib/components/Table/components/TableCell/components/NTagCell/nTagCell";
+import { Link } from "@databiosphere/findable-ui/lib/components/Links/components/Link/link";
+import { BasicCell } from "@databiosphere/findable-ui/lib/components/Table/components/TableCell/components/BasicCell/basicCell";
+import {
+  MetadataValue,
+  NTagCell,
+} from "@databiosphere/findable-ui/lib/components/Table/components/TableCell/components/NTagCell/nTagCell";
 import { ColumnDef } from "@tanstack/react-table";
 import type { JSX } from "react";
 import type {
@@ -21,8 +27,15 @@ import {
   processAggregatedOrArrayValue,
   processEntityValue,
 } from "../apis/azul/hca-dcp/common/utils";
-import * as C from "../components";
-import { MetadataValueTuple } from "../components/common/NTagCell/components/PinnedNTagCell/pinnedNTagCell";
+import { ButtonOutline } from "../components/common/Button/components/ButtonOutline/buttonOutline";
+import {
+  MetadataValueTuple,
+  PinnedNTagCell,
+} from "../components/common/NTagCell/components/PinnedNTagCell/pinnedNTagCell";
+import { BioNetworkCell } from "../components/common/Table/components/Cell/components/BioNetworkCell/bioNetworkCell";
+import { CXGDownloadCell } from "../components/common/Table/components/Cell/components/CXGDownloadCell/cxgDownloadCell";
+import { TrackerDownloadCell } from "../components/common/Table/components/Cell/components/TrackerDownloadCell/trackerDownloadCell";
+import { AnalysisPortalCell } from "../components/HCABioNetworks/Network/Atlas/components/Overview/components/MainColumn/components/AnalysisPortalCell/analysisPortalCell";
 import { NETWORKS_ROUTE } from "../constants/routes";
 import { formatCountSize } from "../utils/formatCountSize";
 import { buildTrackerDownloadCellProps } from "../utils/trackerNetwork";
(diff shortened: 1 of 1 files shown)


NoopDog (approved):
Verified the rename is behavior-preserving: all ten direct imports resolve to the same modules the `components` barrel re-exported, no props dropped or renamed across the 17 conversions, and the `bioNetworkCell` import cycle is unchanged from main with no module-scope TDZ hazard. `tsc --noEmit`, `eslint`, and `prettier --check` all clean on the branch.

Nit, non-blocking: the PR body says `CXGDownloadCell` uses `useDialog()` — it uses plain `useState`; `TrackerDownloadCell` is the one with `useDialog()`. The fix itself is right either way.

PR #3223, opened 2026-09-10
refactor: align the breast atlas path and content folder with the other atlases (#3220)

### LINKED ISSUE #3220: refactor: rename the breast atlas path from breast-v1 to breast-v1-0
## Problem

Breast v1.0 is the only atlas whose path omits the minor version. Every other atlas spells it out:

```ts
// constants/networks.ts
const BRAIN_V1_0 = "brain-v1-0";
const BREAST_V1  = "breast-v1";     // <- the odd one out
const CORTEX_V1_0 = "cortex-v1-0";
const GUT_V1_0 = "gut-v1-0";
const LIVER_V1_0 = "liver-v1-0";
```

So the URLs read inconsistently:

```
/hca-bio-networks/lung/atlases/lung-v1-0
/hca-bio-networks/eye/atlases/retina-v1-0
/hca-bio-networks/breast/atlases/breast-v1        <- should be breast-v1-0
```

The atlas is named "iHBCA v1.0" everywhere else — the display name, the content, the tracker's `version: "v1.0"` — so the path is the only place it appears as bare `v1`.

## Changes

- `constants/networks.ts:40` — `const BREAST_V1 = "breast-v1"` becomes `breast-v1-0`. The constant is used for both `key` and `path` (lines 77 and 79) plus the content lookup at line 675, so one edit moves all three. Worth renaming the constant to `BREAST_V1_0` to match its siblings.
- `@types/network.ts` — the `AtlasKey` union member `"breast-v1"` becomes `"breast-v1-0"`. `tsc` will flag every remaining reference, so the union is the safety net for this rename.

Note the atlas `key` doubles as the content-map key (`NETWORK_ATLAS_CONTENT`), so changing it also changes which entry the content resolves under — check the MDX under `content/breast/atlases/breast/v1/` still loads. That directory is keyed by its own path rather than the atlas key, so it likely needs no change, but it should be verified rather than assumed.

## URLs affected

Two pages move:

```
/hca-bio-networks/breast/atlases/breast-v1                  -> …/breast-v1-0
/hca-bio-networks/breast/atlases/breast-v1/source-datasets  -> …/breast-v1-0/source-datasets
```

## No redirect

Accepted that both old URLs 404 after this lands.

The site is a static export (`output: "export"` in `next.config.mjs`), so Next `redirects()` are unavailable and any redirect would have to come from CloudFront or S3. Not warranted here: these URLs are reached through in-app navigation, and the same reasoning was applied to the two URL changes in epic #3193 (#3194 and #3198).

## Acceptance criteria

- [ ] The atlas serves at `/hca-bio-networks/breast/atlases/breast-v1-0` and `/hca-bio-networks/breast/atlases/breast-v1-0/source-datasets`.
- [ ] Neither `breast-v1` path is emitted into the static export any more - `find out -path "*breast-v1*"` matches only the new spelling.
- [ ] The generated sitemap lists the new paths and no longer lists the old ones.
- [ ] The atlas page renders unchanged: overview content, the source datasets table with its seven rows, all filters, and the Primary Data and Explore links.
- [ ] No `breast-v1` string remains outside the new path - check `constants/networks.ts`, `@types/network.ts`, and content.

## Notes

Cosmetic and low risk, but it is the kind of inconsistency that is cheapest to fix while only one atlas has it - every tracker-sourced atlas added from here on will follow whichever convention is in place.

Compare the generated output by content rather than file hashes - this build is not byte-deterministic between runs.

Do this before #3199's Primary Data work lands anywhere that hardcodes the path, and note that the tracker lookup is by `shortNameSlug`/`version` (`"breast"` / `"v1.0"`), not by this path, so the tracker integration is unaffected.

frano-m:
## Summary

Closes #3220.

Breast v1.0 was the only atlas whose path omitted the minor version, and — spotted while implementing — the only one whose content folder was `v1` rather than `v1.0`.

| | breast (before) | breast (after) | the other eight |
| --- | --- | --- | --- |
| URL path | `breast-v1` | **`breast-v1-0`** | `lung-v1-0`, `retina-v1-0`, `brain-v1-0`, … |
| Atlas key | `breast-v1` | **`breast-v1-0`** | matches path |
| Content folder | `v1` | **`v1.0`** | all eight are `v1.0` |
| `version` field | `"v1"` | `"v1"` | all nine are `"v1"` — already aligned |

Note the two conventions differ deliberately: the URL path uses dashes (`breast-v1-0`), the content folder uses a dot (`v1.0`), matching the sibling folders.

## Changes

- `constants/networks.ts` — `BREAST_V1` becomes `BREAST_V1_0 = "breast-v1-0"`. That one constant supplies the atlas `key`, the `path` *and* the `NETWORK_ATLAS_CONTENT` map key, so all three move together and cannot drift.
- `@types/network.ts` — the `AtlasKey` union member becomes `"breast-v1-0"`, still in alphabetical position. This is the safety net: `tsc` flags any missed reference.
- `content/breast/atlases/breast/v1` → `v1.0`, with the import in that folder's `index.ts` updated to match.

## URLs affected — no redirect

```
/hca-bio-networks/breast/atlases/breast-v1                  -> …/breast-v1-0
/hca-bio-networks/breast/atlases/breast-v1/source-datasets  -> …/breast-v1-0/source-datasets
```

Both old URLs 404, as #3220 specifies. `output: "export"` means Next `redirects()` are unavailable and any redirect would need CloudFront or S3 — the same reasoning applied to the two URL moves in epic #3193 (#3194, #3198).

## Verified rather than assumed

The three things that could have broken, all checked:

- **Content still resolves.** The content module is imported by *path* (`../content/breast/atlases/breast`), and looked up at runtime as `NETWORK_ATLAS_CONTENT[networkKey]?.[key]` — both sides use the renamed constant. Confirmed by the MDX description rendering on the built page after the folder rename.
- **The tracker integration is untouched.** It resolves the atlas by `shortNameSlug: "breast"` / `version: "v1.0"`, not by this path, so the API lookup is unaffected. The source datasets page builds and still shows its seven rows.
- **No stale references.** `grep -rn '"breast-v1"'` returns nothing, and only `breast-v1-0` paths appear in `out/` — the old path is emitted nowhere.

## Two notes

**One transient build failure, not reproducible.** The first build after the rename failed prerendering `/hca-bio-networks/eye` — a network page unrelated to anything here. Two subsequent clean builds passed with no fetch warnings, so it looks like a transient external API failure during static generation. Flagging it because it is the same class of build fragility as #3203; if it recurs on CI it is worth chasing there rather than treating as noise.

**`inclusion.mdx` is not part of this.** Breast has no `inclusion.mdx`, but nor do gut, liver, or either organoid — five of nine, including breast's tracker-sourced peers. That is content presence rather than a naming convention, so it is left alone.

## Interaction with #3222

None. #3222 changes the breast `code` link URL; this touches the path constant, the atlas key and the content folder. Different lines of `constants/networks.ts`, so either can land first.


NoopDog (approved):
Reviewed the rename of `breast-v1` → `breast-v1-0` and the `v1/` → `v1.0/` content folder move.

- No stale references: `grep -rn "breast-v1"` returns only the two intended definition sites.
- `key`, `path`, and the `NETWORK_ATLAS_CONTENT` map key all derive from the single `BREAST_V1_0` constant, so routing (`atlas.path`) and content resolution (`atlas.key`) can't drift.
- Tracker lookup is unaffected — it matches on `shortNameSlug` + `tracker.version`, and `Atlas.version` stays `"v1"` like the other eight atlases.
- `npx tsc --noEmit` is clean; the `AtlasKey` union edit is the safety net it claims to be.

Two intended behavior changes, noted for the record: the old `/breast-v1` URLs will 404 with no redirect (`output: "export"` rules out Next `redirects()`, same as #3194 and #3198), and the Hero breadcrumb renders the raw `atlas.key` so it now reads `breast-v1-0`.

LGTM.

PR #3224, opened 2026-09-10
refactor: extract the source study citation into its own component (#3218)

### LINKED ISSUE #3218: refactor: FileNameCell guards the citation on the split result rather than on publicationString
## Problem

Raised in review of #3213, which merged. Nothing here is broken — this is a readability question about the shape the cleanup left behind, filed so it is tracked rather than living only in a PR thread.

#3213 hoisted the `splitTrailingWord` call out of the JSX so the citation is split once per row:

```tsx
// components/.../FileNameCell/fileNameCell.tsx
const citation = publicationString
  ? splitTrailingWord(publicationString)
  : null;
…
{citation && (
  <Typography …>
    {doi ? (
      <Link label={<>{citation.head}<StyledNoWrap>{citation.tail}…</StyledNoWrap></>} … />
    ) : (
      publicationString
    )}
  </Typography>
)}
```

That is correct — `splitTrailingWord` returns an object literal on both paths, so `citation` is non-null exactly when `publicationString` is a non-empty string, and `publicationString: string | null` (`@types/network.ts:247`) has no other falsy-but-meaningful value. Three things about the result are worth a second look:

1. **The guard is indirect.** `{citation && …}` reads as "if we have a citation", but the block it guards renders bare `publicationString` in the no-DOI branch. Confirming the two conditions coincide means going and reading `splitTrailingWord`'s return type.

2. **Narrowing is lost.** Because the guard is on `citation` rather than on `publicationString`, TypeScript no longer narrows `publicationString` to `string` inside the block, so the no-DOI branch renders `string | null`. React renders `null` as nothing and `tsc` is happy, so this is cosmetic — but it is the kind of thing that makes a future reader stop.

3. **The split runs when it isn't used.** For a row with a `publicationString` but no `doi`, `citation` is built and discarded. One `lastIndexOf` and two `slice`s per row is nothing; noting it only because it is the flip side of hoisting the call out of the branch that consumes it.

## Approach

Fran's call — including the option of leaving it exactly as it is.

The obvious in-place fix trades one wart for another:

```tsx
{publicationString && (
  …
  {doi ? (
    <Link label={<>{citation?.head}…
```

which restores the direct guard and the narrowing, but reintroduces a `?.` on a value that cannot be null at that point.

If it is worth more than that, the version that addresses all three is a small subcomponent taking `publicationString: string` — the split then lives next to its only consumer, the guard stays on `publicationString`, and the no-DOI branch is a plain string. That is more machinery than five lines of JSX obviously justifies, which is why this is discretionary rather than specified.

## Acceptance criteria

Only if the issue is taken up:

- [ ] Rendered output is unchanged — citation present iff `publicationString` is non-empty; DOI rows link the whole citation, with the final word and the external-link icon held on one line; no-DOI rows render plain text.
- [ ] The citation is still split at most once per row.

## Notes

**Update or close at your discretion** — there is no bug and no user-visible symptom. Closing this as "working as intended" is a perfectly good outcome; it is filed so the observation isn't lost, not because it needs doing.

Context: #3213 was itself a follow-up cleanup to #3196 (#3209), part of #3193.

frano-m:
## Summary

Closes #3218.

`FileNameCell` guarded its citation line on the *result* of the split rather than on the citation itself. #3218 raised three consequences and offered the shape that fixes all three: a subcomponent taking a required `publicationString: string`.

```tsx
// before
const citation = publicationString ? splitTrailingWord(publicationString) : null;
{citation && ( … {doi ? <Link label={…citation.head…} /> : publicationString} … )}

// after
{publicationString && ( … <Citation doi={doi} publicationString={publicationString} /> … )}
```

## The three points, and how each is now closed

| # | Before | After |
| --- | --- | --- |
| 1. Indirect guard | `{citation && …}` guarded a block rendering bare `publicationString` | guard is `{publicationString && …}` — the thing itself |
| 2. Narrowing lost | `publicationString` stayed `string \| null` inside the guard | narrows to `string`, and is passed as a required prop |
| 3. Split ran unused | built and discarded on rows with a citation but no DOI | `if (!doi) return …` sits **above** the split |

**Point 2 demonstrated** rather than asserted: changing the call site to `publicationString.trimEnd()` now compiles with no null assertion, where before it produced `TS18047: 'publicationString' is possibly 'null'`.

The split is still called **once** per render, so #3213's fix is preserved rather than undone.

## Structure

The citation-only pieces moved with it, so the concern is self-contained and `fileNameCell.tsx` is back to a `Stack` plus two lines:

```
FileNameCell/
  fileNameCell.tsx              Stack + file name + <Citation/>
  types.ts
  components/Citation/
    citation.tsx                takes publicationString: string
    citation.styles.ts          StyledNoWrap, StyledOpenInNewIcon   (moved)
    constants.ts                DOI_BASE_URL, EXTERNAL_LINK_TITLE   (moved)
    types.ts
    utils.ts                    splitTrailingWord                    (moved)
```

`citation.styles.ts` and `constants.ts` are 100%-similarity renames, so history follows them. `utils.ts` is a rename **with edits** — see *Behaviour change in the moved `splitTrailingWord`* below; it is the one hunk in this PR carrying behaviour, so don't skim it. Folder naming matches the sibling `AnalysisPortalCell` convention.

## Verification

Rendered output is unchanged:

- **7 DOI anchors, all 7 carrying the icon** inside the anchor
- **7 nowrap spans**, each holding the citation's final word plus its `<svg>`
- **7 `(opens in a new tab)`** titles — the `titleAccess` a11y cue survives
- `href=""` count **0**
- citation still `var(--palette-ink-light)`; icon still **16px**
- 7 file names on the primary line, 6 `HCA Explorer` links
- `npm run lint` (0 errors), `npm run check-format`, `npx tsc --noEmit`, `npm run build-prod:data-portal` (64/64) all pass

One equivalence worth spelling out, raised in review: for `publicationString === ""` the old guard produced `null` and rendered nothing; the new guard is falsy and also renders nothing. Equivalent for every value of `string | null`.

## Behaviour change in the moved `splitTrailingWord`

The move was **not** clean — two defects were fixed in passing, so `utils.ts` is the one file here where the diff matters:

**1. Trailing whitespace leaked onto the tail** (36425959). The index was computed from `trimEnd()` but the slices were taken from the original `text`, so `"… Cell   "` put the spaces on the tail and rendered them as a gap between the final word and the icon. Both slices now come from `trimmed`, so the postcondition is `head + tail === text.trimEnd()` — not `text`.

**2. Only U+0020 counted as a separator** (29f05961, raised in review). `lastIndexOf(" ")` found nothing in a citation separated by non-breaking spaces or tabs, returning `head === ""` — which puts the *whole* citation inside the nowrap span and clips the pinned column instead of wrapping it. Now `/\s(?=\S*$)/`:

| input | before | after |
| --- | --- | --- |
| `Gray (2022) Developmental Cell` | `tail="Cell"` | `tail="Cell"` (unchanged) |
| `Gray (2022) Developmental\u00a0Cell` | `head=""` → clips | `tail="Cell"` |
| `Gray (2022)\tDevelopmental\tCell` | `head=""` → clips | `tail="Cell"` |

All 7 real citations in the build are separated by U+0020 and are byte-identical before and after, so this is a latent-input fix, not a visible one.

## Not done

`splitTrailingWord` is a pure function with edge cases that deserve unit tests — single word, trailing whitespace, NBSP/tab separators, empty string — and has none. There is still no test tooling in the repo (#3190). Given it has now changed twice, it is worth seeding first alongside `buildVersionedFileName` and `splitFileName` when that lands.

## No conflict with the open PRs

Checked before starting: this touches only `FileNameCell/`. #3223 touches `@types/network.ts`, `constants/networks.ts` and two `content/breast/` files; #3216 touches `viewModelBuilders.tsx`. No overlap either way, so this sits on `main` rather than stacking.


NoopDog (approved):
Approving — the extraction is clean, callers all resolve, `tsc --noEmit` and `npm run lint` pass, and the rendered output is unchanged for the current data.

Two low-priority notes, neither blocking:

**1. `Citation/utils.ts` — `lastIndexOf(" ")` only matches U+0020.**
A citation whose word separators are non-breaking spaces, tabs, or newlines (plausible for a `publicationString` pasted from a typeset journal page) returns `-1`, so `head` is `""` and the whole citation lands inside `StyledNoWrap` (`white-space: nowrap`). In the pinned Source Datasets column that overflows/clips the cell instead of wrapping to a second line — the opposite of what the helper exists to prevent. Pre-existing behaviour, but since the line was rewritten here it's cheap to close: match any whitespace, e.g. `/\s(?=\S*$)/`.

**2. The PR description undersells one hunk.**
The "Not done" section says `splitTrailingWord` "moved unchanged", but 36425959 slices `tail` from `text.trimEnd()`, so `head + tail === text.trimEnd()` rather than `text`. The change is correct and a genuine improvement — a trailing space inside a `nowrap` span renders as a visible gap before the icon — it just isn't part of the 100%-similarity renames, and a reviewer skimming that claim would skip the one hunk with behaviour in it.

https://claude.ai/code/session_018z2yTzwXdsj9qfjxzTHTk7

PR #3225, opened 2026-09-10
fix: make icon marks decorative and announce links that open a new tab (#3212)

### LINKED ISSUE #3212: fix: analysis portal links give no 'opens in a new tab' cue and duplicate their icon's alt text

frano-m:
## Summary

Closes #3212.

Two accessibility defects around the portal marks, plus the same redundant-`alt` pattern in three more components.

**Before**, `IconLink` rendered the mark and the anchor as siblings, both labelled with the same string, and the anchor opened a new tab with no signal:

```html
<img alt="CAP" height="20" src="…/cap.svg"/>
<a href="https://celltype.info/…" rel="noopener noreferrer" target="_blank">CAP</a>
```

so a screen reader announced "CAP, image" then "CAP, link", and nothing said the link leaves the site.

## Changes

**`components/common/IconLink/iconLink.tsx`**

- The mark is now decorative (`alt=""`) — the adjacent link text already names the destination.
- Links that open a new tab append a visually-hidden cue to their accessible name.

**The cue lives in one shared component.** `NewTabCue` renders the visually-hidden text and owns the decision, so `IconLink` delegates rather than computing it — leaving `IconLink` simpler than it was before this branch.

**The cue is derived, not asserted.** `IconLink` has two consumers with different link kinds, so an unconditional cue would have been wrong:

| Consumer | URL | Target |
| --- | --- | --- |
| `AnalysisPortalCell` | `https://celltype.info/…` | `_blank` |
| `HCABiologicalNetwork` | `/hca-bio-networks/breast` | same tab |

Rather than have callers declare it, the component uses the same two predicates `Link` uses to choose its target:

```ts
const opensInNewTab = !isClientSideNavigation(url) && isValidUrl(url);
```

so the cue cannot drift from where the link actually opens.

**Three more components with the same redundant `alt`** — `sectionBioNetworkAtlases`, `bioNetworkCell` and `heroTitle` each rendered the network icon with `alt` set to the text sitting right beside it. None is a portal link (one is a click handler, one an internal link, one a heading), so only the decorative-`alt` half applies to them. These are beyond #3212's stated scope, included deliberately as the same defect.

**Standalone images keep their alt text** — verified across the export, the only images still carrying `alt` are the header and footer logos (63 pages each), the metadata-schema graphic, and the trademark mosaic ball on `/about/attributions`. None sits beside text repeating it.

~~the BICCN … graphic~~ — **this was wrong, and is now fixed.** `Tabs/tabs.tsx:54` rendered `alt="BICCN"` on the icon of a tab labelled `"BICCN Publications"`, announcing *"BICCN BICCN Publications"* — the exact defect this PR exists to fix, caught in review. Now `alt=""`; `alt="BICCN"` is gone from the export (2 → 0).

## Verification

Audited every icon image and every anchor in the generated export:

```
network/portal <img> with alt=""   232
cues wrongly on same-tab links       0
cues missing on _blank links         0   <- on the surfaces this PR touches
```

**Scoping that last line**, which was over-claimed: it holds for the links this PR touches, not site-wide. Three explicit-`_blank` links elsewhere still announce nothing — `viewModelBuilders.tsx:509`, `sectionHero.tsx:41` and `sectionStayUpToDate.tsx:31` — plus the social and footer entries in `site-config/data-portal/dev/`. So the cue is currently inconsistent site-wide.

(The review listed a fourth, `fileNameCell.tsx:60`. That one — now `Citation/citation.tsx:44` after #3224 moved it — already announces, via `titleAccess` on its external-link icon rather than a hidden span. Three, not four.)

Every anchor carrying the cue, by link text — all `_blank`, none same-tab:

```
'CZ CELLxGENE'      x167      'CAP'                x2
'HCA Explorer'        x6      'UCSC Cell Browser'  x1
7 source study citations (via #3196's titleAccess)
```

The two `IconLink` cases on the atlas overview page, side by side:

```
alt=''  text='CAP'             target=_blank    new-tab cue=YES
alt=''  text='Breast Network'  target=same tab  new-tab cue=no
```

Surfaces exercised: the source datasets table's Explore column (CAP), the atlas overview integrated-objects table (CAP, CZ CELLxGENE ×167, UCSC Cell Browser), the network link in the atlas overview side column, the home page network cards, the network table cells, and the atlas hero.

## The Primary Data link is included

#3212 listed the HCA Data Explorer link in the Primary Data column, and half of that is now stale: #3199 shipped it as text only, with no mark, so it never had the duplicate-`alt` problem.

It *did* still open in a new tab announcing nothing, because it renders a plain `Link` rather than going through `IconLink`. Sharing the cue as `NewTabCue` made fixing that a one-line change rather than a duplicated span, so it is in this PR — the six `HCA Explorer` links now carry the cue, verified above.


components/common/NewTabCue/newTabCue.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * inside the link's label so it forms part of the accessible name. Decides
+ * from the same predicates findable-ui's `Link` uses to choose its target -
+ * internal URLs stay in the same tab and get nothing - so the cue cannot
+ * disagree with where the link actually opens.
+ * @param props - Component props.
+ * @param props.url - The link's URL.
+ * @returns the cue, or null when the link stays in the same tab.
+ */
+export const NewTabCue = ({ url }: Props): JSX.Element | null => {
+  if (isClientSideNavigation(url) || !isValidUrl(url)) return null;
```
NoopDog, line 19: **Medium.** The cue is derived from the URL alone and ignores an explicit `target`, while findable-ui's `Link` lets `target` override its default (`target || ANCHOR_TARGET.BLANK` / `SELF`).

No live defect today - neither `IconLink` nor the primary-data `Link` passes `target`. But the repo already has four `Link`s that set `target={ANCHOR_TARGET.BLANK}` explicitly (`viewModelBuilders.tsx:509`, `sectionHero.tsx:41`, `sectionStayUpToDate.tsx:31`, `fileNameCell.tsx:60`), and the social/footer config adds more. The next maintainer reusing `NewTabCue` on an internal URL forced to `_blank` gets silence, and on an external URL forced to `_self` gets a cue that lies about where the link opens.

An optional `target` prop would close the gap: `BLANK` -> cue, `SELF` -> null, otherwise derive as now.

components/common/NewTabCue/constants.ts
```diff
@@ -0,0 +1,3 @@
+// Appended to a link's accessible name when it opens in a new tab. Leading
+// space so it reads as a continuation of the label it follows.
+export const NEW_TAB_LABEL = " (opens in a new tab)";
```
NoopDog, line 3: **Low.** This duplicates the existing `EXTERNAL_LINK_TITLE = "(opens in a new tab)"` at `components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/components/FileNameCell/constants.ts:6`.

Two constants for one user-facing string, rendered on the same pages, will drift the moment the wording is revised. This new shared module is the natural home for both - `FileNameCell` can import the text from here (its `titleAccess` wants the variant without the leading space, so export both, or trim at the use site).

components/common/NewTabCue/newTabCue.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return <span style={visuallyHidden}>{NEW_TAB_LABEL}</span>;
```
NoopDog, line 20: **Low.** `style={visuallyHidden}` is the only inline `style` prop in `components/`, `views/` and `pages/` - a grep for ` style={` across all three returns nothing else. The repo consistently uses a sibling `*.styles.ts` styled component instead.

A `StyledVisuallyHidden` span in `newTabCue.styles.ts` (spreading `visuallyHidden`) would match convention and keep the style overridable by consumers.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
```diff
@@ -77,10 +78,16 @@ export function buildRenderPrimaryData(
   return function renderPrimaryData(ctx) {
     const { hcaProjectId } = ctx.row.original;
     if (!hcaProjectId) return null;
+    const url = buildHCADataExplorerProjectUrl(browserUrl, hcaProjectId);
```
NoopDog, line 81: **Low.** (Anchored here; the text in question is lines 69-70, just outside this hunk.) The `buildRenderPrimaryData` JSDoc still says "Text only - `Link` sends external URLs to a new tab with noopener/noreferrer", which this change makes false: the label is no longer text only, it now carries the `NewTabCue` element. Worth rewording to say the label carries the visually-hidden new-tab cue.

NoopDog (commented):
Code review (high effort). No correctness bug found: the `NewTabCue` predicate is an exact match for findable-ui `Link`'s branch order (`isURLObjectWithHrefAndQuery` -> client-side -> `isValidUrl` -> plain `Typography`), including the invalid/empty-URL case where `Link` renders no anchor and the cue correctly renders nothing; `label: ReactNode` accepts the fragments; `StaticImageProps.alt` is a plain required `string`, so `alt=""` type-checks and emits a correctly decorative ``; `visuallyHidden` is a real export of the installed `@mui/utils@7.3.11`; and both `IconLink` consumers keep visible link text (`StyledIconLink` only overrides `font`), so no link is left unnamed by the `alt=""` change.

Five findings, all quality-level. Four are inline. Two items could not be anchored to a diff line:

**Low - `components/HCABioNetworks/Network/components/common/Tabs/tabs.tsx:53`** (file not in this PR, so not inline): a missed instance of exactly the defect this PR fixes. `<StaticImage alt="BICCN" ... />` is the `icon` of a tab whose `label` is `"BICCN Publications"`, so the tab announces "BICCN BICCN Publications". The PR description asserts the audit found the BICCN graphic has no adjacent text repeating it; this one does.

**On the verification claim** (not line-anchorable): "cues missing on `_blank` links 0" holds only for the surfaces touched here. Four explicit-`_blank` `Link`s still announce nothing - `viewModelBuilders.tsx:509` (project title), `components/Home/components/SectionHero/sectionHero.tsx:41`, `components/Home/components/SectionStayUpToDate/sectionStayUpToDate.tsx:31`, and `fileNameCell.tsx:60` - plus the social-media and footer entries in `site-config/data-portal/dev/`. So the cue is currently inconsistent site-wide; worth either scoping that claim in the description or filing a follow-up.

components/common/NewTabCue/newTabCue.tsx
frano-m: Fixed in `e52e8518`.

`NewTabCue` now mirrors `Link`'s own precedence rather than deriving from the URL alone:

```tsx
if (target === ANCHOR_TARGET.SELF) return null;
if (target !== ANCHOR_TARGET.BLANK) {
  if (isClientSideNavigation(url) || !isValidUrl(url)) return null;
}
return <StyledSpan>{NEW_TAB_LABEL}</StyledSpan>;
```

So `BLANK` → cue, `SELF` → null, and absent an explicit target the URL decides exactly as before. `target` is optional on `Props`, so both existing call sites are unchanged.

Agreed there is no live defect today — neither `IconLink` nor the primary-data `Link` passes `target`. This is for the next consumer, and both of the ways it could have lied (internal URL forced to `_blank` → silence; external forced to `_self` → a cue that misstates where the link opens) are now closed.

components/common/NewTabCue/constants.ts
frano-m: Fixed in `e52e8518`, and this module is now the single home for the wording as you suggested.

```ts
// components/common/NewTabCue/constants.ts
export const NEW_TAB_TEXT  = "(opens in a new tab)";   // bare, for titleAccess
export const NEW_TAB_LABEL = ` ${NEW_TAB_TEXT}`;       // leading space, for a label
```

Rather than exporting two independent literals I derived the label from the bare text, so a revision to the wording cannot half-land. `EXTERNAL_LINK_TITLE` is deleted and `Citation` imports `NEW_TAB_TEXT` for its icon `titleAccess`; a repo-wide grep for `EXTERNAL_LINK_TITLE` now returns nothing, and `"opens in a new tab"` is defined in exactly one place.

One correction: the constant had already moved in #3224, so it was at `FileNameCell/components/Citation/constants.ts:6` rather than `FileNameCell/constants.ts:6`. Same finding, just a stale path — that file now holds only `DOI_BASE_URL`.

The export count is unchanged in the build: 176 hidden cues and 7 icon `titleAccess` cues before and after.

components/common/NewTabCue/newTabCue.tsx
frano-m: Fixed in `e52e8518`. Your grep was right — it was the only inline `style` in `components/`, `views/` and `pages/`, and it is now zero.

`newTabCue.styles.ts`:

```ts
export const StyledSpan = styled.span(visuallyHidden as CSSObject);
```

Worth flagging one thing for whoever touches this next: the obvious template form does not compile.

```ts
styled.span`${visuallyHidden}`   // TS2769 — CSSProperties is not an Interpolation
```

`visuallyHidden` is typed `CSSProperties`, which emotion will not accept as a template interpolation (no string index signature), so it needs the object call form with the cast. I left that reason in the file so it does not get "simplified" back into a template and break the build.

Confirmed the emitted rule is equivalent to the inline style it replaces, rather than assuming:

```
.mui-1am6aje{border:0;clip:rect(0 0 0 0);height:1px;margin:-1px;overflow:hidden;
             padding:0;position:absolute;white-space:nowrap;width:1px;}
```

Same nine declarations, and accessible text is unchanged across the export — `'CAP (opens in a new tab)'` before and after. Across the whole build: 176 cues before and after, inline-styled 176 → 0.

components/HCABioNetworks/Network/Atlas/components/SourceDatasets/components/MainColumn/components/table/viewBuilder.tsx
frano-m: Fixed in `e52e8518`. Copilot flagged the same line independently, so it got two votes.

Before:

> ... than through untyped table meta. **Text only** - `Link` sends external URLs to a new tab with noopener/noreferrer.

After:

> ... than through untyped table meta. **The label carries a visually-hidden new-tab cue alongside the text**, because `Link` sends external URLs to a new tab with noopener/noreferrer.

The "text only" claim was the misleading part, since the accessible name is the thing a caller would reason about and it is no longer just `HCA_EXPLORER_LABEL`. Thanks for anchoring it despite the text sitting outside the hunk — it would have been easy to leave stale.

NoopDog (approved):

PR #3231, opened 2026-09-14
refactor: build site config branding and icons as jsx, not eager calls (#3217)

### LINKED ISSUE #3217: refactor: site config builds React elements eagerly by calling components as functions
## Problem

`site-config/data-portal/dev/config.ts` builds elements by invoking components as plain functions, at module scope:

```ts
footer: {
  Branding: C.Logo({ alt: APP_TITLE, height: 38, link: "…" }),
},
header: {
  logo: C.Logo({ alt: APP_TITLE, height: 32, … }),
},
…
{ icon: C.GitHubIcon({ fontSize: "small" }) },
{ icon: C.FacebookIcon({ fontSize: "small" }) },
{ icon: C.XIcon({ fontSize: "small" }) },
{ icon: C.LinkedInIcon({ fontSize: "small" }) },
```

Six call sites. Because this is a config module rather than a component, the calls run **when the module is first imported**, not during a render — so each component body executes once, outside React's rendering cycle, and the resulting element is stored in the config object.

## Why it is not urgent

None of the five components involved (`Logo`, `GitHubIcon`, `FacebookIcon`, `XIcon`, `LinkedInIcon`) uses hooks, so there is no hook-ordering hazard and nothing is broken today. That is the distinction from #3211, where `TrackerDownloadCell` and `CXGDownloadCell` both call `useDialog()` and were genuinely at risk.

## Why it is still worth deciding

- **The invariant is invisible.** Nothing stops someone adding a hook, `useTheme`, or context read to `Logo` — at which point it executes outside a render with no component instance, and fails in a way that points at the config file rather than the component.
- **Elements built at import time cannot respond to anything.** They are frozen at module evaluation, so they cannot read theme, locale, or any provider. If the branding ever needs to vary, this shape has to be unpicked first.
- **It is inconsistent** with how cells are now built after #3211, and with `.tsx` config being available.

## Options

1. **Pass component references and props instead of elements** — `Branding: Logo` plus a props object, letting the consumer render it. Cleanest, but depends on what findable-ui's `Header`/`Footer` accept for these fields.
2. **Rename the config to `.tsx` and use JSX** — `Branding: <Logo … />`. Still builds the element eagerly, but at least states what it is and keeps the door open for a render-time wrapper later.
3. **Leave it, and document the constraint** — a comment stating that components referenced from site config must stay hook-free and render-independent.

Worth checking what those findable-ui props are typed as before choosing; if they are `ReactNode`, option 1 may not be available without a library change.

## Scope

Also audit the other four modules still importing the barrel, for the same pattern:

- `mdx-components.tsx`
- `docs/common/constants.ts`
- `site-config/data-portal/dev/socialMedia.ts`
- `site-config/data-portal/dev/layout/floating.ts`

A repo-wide sweep for non-barrel direct component calls (`PascalCase({ … })` in `.ts`, `return PascalCase({ … })` in `.tsx`) came back clean, so these six plus whatever the four modules above hold should be the complete set.

## Acceptance criteria

- [ ] A decision is recorded for how site config supplies branding and icon elements.
- [ ] No component is invoked as a plain function in `site-config/`, or the constraint is documented where someone editing `Logo` would see it.
- [ ] The header, footer, floating elements and social-media icons render unchanged - compare the generated `out/` content, not file hashes (this build is not byte-deterministic between runs).

## Notes

Found while auditing for the pattern #3211 fixes. Kept separate because the hazard is different in kind: #3211 was about hooks inside table cell renderers during render; this is about elements constructed at import time in configuration.

frano-m:
## Summary

Closes #3217.

`site-config/data-portal/dev/config.ts` built six elements by calling components as plain functions. `makeConfig` runs at module evaluation (line 217, and again from `prod/config.ts`), so each component body executed **outside React's rendering cycle** with no component instance.

The file is now `.tsx` and the six sites are JSX.

```tsx
// before                                       // after
Branding: C.Logo({ alt: APP_TITLE, … })         Branding: <Logo alt={APP_TITLE} … />
logo:     C.Logo({ alt: APP_TITLE, … })         logo:     <Logo alt={APP_TITLE} … />
icon:     C.GitHubIcon({ fontSize: "small" })   icon:     <GitHubIcon fontSize="small" />
```

## Which option, and why

The issue left the choice open pending findable-ui's prop types, so I read them first:

| field | type in findable-ui | takes |
| --- | --- | --- |
| `HeaderProps.logo` | `ReactNode` | an element |
| `FooterProps.Branding` | `ReactNode` | an element |
| `MenuItem.icon` | `ReactNode` | an element |
| `Social.Icon` | `ElementType` | a **reference** |
| `ComponentConfig.component` | `ElementType` | a **reference** |

All three fields in question are `ReactNode`, so **option 1 (pass references + props) is not available without a library change** — exactly the branch the issue anticipated. That leaves option 2, the `.tsx` rename.

Worth correcting one line in the issue, though: option 2 is described as *"still builds the element eagerly"*. The **descriptor** is still built eagerly, but the **component body is not executed** — which is the entire hazard. Verified rather than assumed:

```
A  Logo({ … })                      body executed: true   result: the component's return value
B  jsx(Logo, { … })                 body executed: false  result: descriptor, type = Logo, props = {…}
   (react/jsx-runtime, i.e. what
    `jsx: "react-jsx"` actually emits)
```

So `<Logo … />` closes the hook hazard outright; it is not merely a cosmetic restatement. (Copilot correctly flagged an earlier draft of this that said "compiles to `createElement`" — `tsconfig.json` sets `jsx: "react-jsx"`, so it emits `_jsx()` from `react/jsx-runtime`. Different function, same observable behaviour, re-proved above against the runtime actually in use.) A hook added to `Logo` tomorrow now runs under the reconciler, where it belongs.

## Audit of the other four modules

All clean — each already passes a **reference**, not a call:

| module | pattern | verdict |
| --- | --- | --- |
| `socialMedia.ts` | `Icon: C.GitHubIcon` | reference — correct already |
| `layout/floating.ts` | `component: C.CookieBanner` | reference — correct already |
| `docs/common/constants.ts` | `Alert: C.Alert`, `Link: C.Link`, … | references — correct already |
| `mdx-components.tsx` | `Link: C.Link` | reference — correct already |

A repo-wide sweep for `C.PascalCase(` now returns **zero** hits, and `PascalCase({` in `.ts` under `site-config/` and `docs/` returns none. `config.ts` was the only offender, which matches the issue's finding.

`prod/config.ts` needed no change — it delegates to dev's `makeConfig` and holds no component calls of its own.

## The decision is recorded in the file

Acceptance criterion 1 asks for the decision to be written down where someone editing `Logo` would meet it, so `config.tsx` now opens with a note stating why the file is `.tsx`, which findable-ui fields take elements vs references, and why JSX rather than a call. It cites #3217.

## Verification

Built `main` first to capture a **pre-change baseline**, then rebuilt with the change and compared content — not file hashes, since this build is not byte-deterministic (the build ID and `_app` chunk hash differ between any two runs, as the issue warns).

- **63/63 HTML pages byte-identical** after normalising the build ID
- Identical file inventory apart from the build-ID-named directories
- Elements positively present, so "identical" isn't two empty renders: header `logoHca.png` ×2, footer `logoHumanCellAtlas.png` ×2, `humancellatlas.org` branding link, all four social URLs, and **4 `MuiSvgIcon-fontSizeSmall`** icons — the `fontSize="small"` prop survives the conversion
- `npm run lint` (0 errors, same 8 pre-existing warnings), `npm run check-format`, `npx tsc --noEmit`, `npm run build-prod:data-portal` (64/64) all pass

Git recorded the move as a rename at 79% similarity, so history follows the file.

## Second commit: the barrel

`6a151249` drops `import * as C from "../../../components/index"` in favour of five direct imports. Kept as its own commit so the JSX conversion in `d2d96f6c` stays readable on its own.

All five components (`Logo` and the four social icons) are plain re-exports from findable-ui, so the barrel was pure indirection here — the direct imports point at findable-ui, not at local code.

**This is a convention change, not a bundle win.** Measured both builds rather than assuming:

| | chunks | total JS |
| --- | --- | --- |
| barrel | 24 | 1,805,175 bytes |
| direct | 24 | 1,805,601 bytes |

**+426 bytes (0.02%), i.e. marginally larger.** The barrel was already being tree-shaken, so anyone expecting this to trim the bundle should not. The reason to do it is that `<Logo />` names what it renders.

## A note on verifying "output unchanged" here

The first comparison ran with the change uncommitted, so `HEAD` was still `main`. Once committed, all 63 pages appeared to differ — the build embeds the **git short SHA** in the footer version chip (`Unversioned-31003d8-dcp60` → `Unversioned-d2d96f6-dcp60`), and the `[...slug]` chunk hash shifts with the import graph.

Neither is content. With the build ID, chunk hashes and embedded SHA normalised, it is **63/63 identical** again — and the chunk *names* are compared unnormalised, so a page silently loading a different chunk would still show up. Worth knowing for the next person who diffs `out/` on this repo: the build is non-deterministic **and** SHA-stamped.


NoopDog (approved):
Reviewed: no correctness issues found. The direct findable-ui imports resolve to the same modules the old barrel re-exported, the config.ts → config.tsx rename resolves for both importers (config/config.ts and site-config/data-portal/prod/config.ts), and each call-to-JSX conversion keeps the same props. LGTM.

PR #3233, opened 2026-09-17
refactor: require versionedfilename on tracker-built dataset assets (#3207)

### LINKED ISSUE #3207: refactor: require versionedFileName on tracker-built dataset assets so a missing name can't silently hide the download button
## Problem

`DatasetAsset.versionedFileName` is optional, because CXG assets are built elsewhere and rendered by `CXGDownloadCell`, which does not use it. But both tracker render paths hard-require it:

```ts
// utils/trackerNetwork.ts
export function buildTrackerDownloadCellProps(
  asset: DatasetAsset | undefined
): TrackerDownloadCellProps | null {
  if (!asset?.versionedFileName) return null;   // <- degrades silently
  ...
}
```

A tracker `DatasetAsset` built without `versionedFileName` type-checks fine, and the consequence at render time is that **the download button vanishes from the row with no error** — no console warning, no fallback, no visual hint that a file exists but cannot be offered.

Today the requirement holds by construction: `buildTrackerDatasetAsset` is the only builder of tracker assets and always sets the field. So this is latent rather than live. It becomes reachable via a new tracker endpoint or code path building an asset, a test fixture, or a hand-built literal — none of which the compiler would stop.

## Approach

Enforce the requirement in the type rather than degrading at render time. Options:

- A tracker-specific subtype:
  ```ts
  export interface TrackerDatasetAsset extends DatasetAsset {
    versionedFileName: string;
  }
  ```
  with `buildTrackerDatasetAsset` returning it and `buildTrackerDownloadCellProps` accepting it. `DatasetAsset` stays as-is for CXG.
- Or `Required<Pick<DatasetAsset, "versionedFileName">> & DatasetAsset` at the boundary, if a full interface feels heavy.

Either way `buildTrackerDownloadCellProps` can drop its `versionedFileName` guard, and the `null` return then means only "no asset" rather than conflating that with "asset without a name".

Worth checking while in there whether `AtlasContext.trackerSourceDatasets` and the integrated-atlas `datasetAssets` can carry the narrower type end to end, or whether the narrowing has to happen at the render boundary.

## Acceptance criteria

- [ ] A tracker dataset asset without a published file name is a compile error, not a silently missing download button.
- [ ] `buildTrackerDownloadCellProps` no longer needs to guard on `versionedFileName`.
- [ ] CXG assets are unaffected — `CXGDownloadCell` still works without the field.
- [ ] The source datasets and integrated objects download cells render exactly as before on Breast v1.0.

## Notes

Raised in review of #3206 (#3200), where it was deliberately deferred: #3200 specifies the field as optional, and `buildTrackerDatasetAsset` is the only tracker asset builder so the invariant holds today. Filed so it is tracked rather than living only in that PR's description.

Not urgent — no user-visible bug exists today. Best done alongside any work that adds a second builder of tracker assets.

frano-m:
## Summary

Closes #3207

- Adds `TrackerDatasetAsset`, extending `DatasetAsset` with a required `versionedFileName`. Both tracker asset builders return it, and `versionedFileName` is removed from the shared `DatasetAsset` so CXG code cannot reference it.
- `TrackerSourceDataset.datasetAsset` is now required and typed as `TrackerDatasetAsset`, so the source datasets download renderer and file name accessor drop their null checks and base-name fallback.
- `buildTrackerDownloadCellProps` takes a `TrackerDatasetAsset` and no longer guards or returns `null`.
- The integrated atlases table shares `IntegratedAtlasRow` between CXG and tracker rows, so narrowing there happens at the render boundary via a new `isTrackerDatasetAsset` type guard. `null` from that cell now means only "no asset".

## Checks

- `npx tsc --noEmit`, `npm run lint` and `npm run check-format` pass (lint warnings are pre-existing on `main`).
- `npm run build-prod:data-portal` currently fails on `main` as well: Azul rejects `size=75` on the projects endpoint with "must not be greater than 50". Unrelated to this change; to be tracked separately.


NoopDog (approved):
Reviewed the full diff (5 files, +54/-30) and traced each affected data path end to end. `tsc --noEmit` and `eslint` are both clean on the branch. No correctness findings — each removed guard is safe:

- **`TrackerSourceDataset.datasetAsset` optional → required.** The sole producer, `fetchTrackerSourceDatasetsWithStudyFields` (`utils/trackerAtlasPages.ts`), unconditionally sets `datasetAsset` on every row, and consumers default to `[]`, so the now-unguarded `row.datasetAsset.versionedFileName` in `accessor.ts` can't dereference undefined.
- **Dropped `|| row.baseFileName` fallback.** `buildVersionedFileName` always returns a non-empty string, so the fallback was already unreachable. Sorting and display still share the same function, so they can't drift.
- **`buildTrackerDownloadCellProps` no longer returns `null`.** Both call sites now supply an asset — one via the required field, the other via the new type guard — so behavior is identical for every reachable input.
- **`isTrackerDatasetAsset` key-presence check.** CXG assets built by `buildDatasetAssets` (`utils/network.ts`) never set `versionedFileName`, so the guard discriminates correctly. It accepts `undefined`, preserving the old `null` return for the empty-`datasetAssets` case. A `{ versionedFileName: undefined }` shape that would pass the guard and then throw in `splitFileName` isn't producible: the builder always assigns a string, and `getStaticProps` JSON serialization strips undefined-valued keys.
- **`const [asset] = row.original.datasetAssets`** has the same failure profile as the previous `datasetAssets[0]` — no new deref risk.

One non-blocking nit, pre-existing and not introduced here: the `fileName` comment on `TrackerComponentAtlas` (`@types/network.ts`) says to use `datasetAsset.versionedFileName`, but `TrackerComponentAtlas` has no `datasetAsset` field — the versioned name only lands on `datasetAssets[0]` after `mapTrackerComponentAtlasToIntegratedAtlas`. The identical comment on `TrackerSourceDataset` is accurate. Worth a fix if you're in that file anyway.

LGTM.

PR #3235, opened 2026-09-17
fix: cap the azul projects page size at 50 and paginate atlas project fetches (#3234)

### LINKED ISSUE #3234: fix: azul rejects the projects request because the default page size of 75 exceeds the new cap of 50, failing the build
## Problem

`npm run build-prod:data-portal` fails on `main` while prerendering atlas pages, e.g. `/hca-bio-networks/eye/atlases/retina-v1-0` and `/hca-bio-networks/lung/atlases/lung-v1-0/datasets`:

```
Error [HTTPError]: Request failed with status code 400 Bad Request: GET https://service.azul.data.humancellatlas.org/index/projects?size=75&catalog=dcp60&filters=...
```

Re-issuing the request by hand returns the reason:

```
{"Code":"BadRequestError","Message":"Invalid value for parameter `size`, must not be greater than 50"}
```

Azul now caps `size` at 50. The site config requests 75 via `dataSource.defaultListParams.size` in `site-config/data-portal/dev/config.tsx`, and every projects fetch for atlas pages inherits it. Because the site is a static export, one failing route fails the whole build.

## Approach

- Lower the default list size to 50 or below.
- Check whether any atlas fetches more than 50 project IDs in a single filtered request (the lung atlas request above carries 29), and if so page the request rather than rely on a single call.

## Acceptance criteria

- [ ] `npm run build-prod:data-portal` passes on `main`.
- [ ] Atlas pages that fetch projects render the same project lists as before.

## Notes

Found while verifying #3233 (#3207); unrelated to that change.

frano-m:
## Summary

Closes #3234

- Azul now rejects `size` above 50 on the projects endpoint, which fails the static export on every atlas page that fetches projects. The default list size in the dev site config drops from 75 to 50.
- The atlas projects fetch in `utils/atlasPages.ts` switches from a single page request to `fetchAllEntities`, which follows pagination, so an atlas with more than 50 datasets still gets every project. The largest atlas today has 39.

## Checks

- `npx tsc --noEmit`, `npm run lint` and `npm run check-format` pass.
- `npm run build-prod:data-portal` passes locally with no prerender errors. It fails on `main` without this change.


NoopDog (approved):
Approved — verified the premise against the live Azul API: `size=51` returns `400 BadRequestError — "Invalid value for parameter \`size\`, must not be greater than 50"`, so `size: "75"` was breaking the build on every atlas page. The `fetchEntitiesFromQuery` → `fetchAllEntities` argument reorder is correct (filters land in `listParams`, catalog still flows in via `defaultParams`), and Azul's `pagination.next` carries a `search_after` tuple tie-broken by entity uid, so paging can't drop or duplicate hits. Every atlas today is a single-hit filtered query, so behaviour is byte-identical to before.

One latent issue I noticed while reading `utils/atlasPages.ts` — **not introduced by this PR**, so no change needed here: the projects block at line 87 (including the `externalDatasets` append and `sortDatasets`) is gated on `atlas.datasets.length > 0`. An atlas configured with an empty `datasets` array but non-empty `externalDatasets` would render an empty dataset list. It's dormant — every atlas in `constants/networks.ts` with external datasets also has non-empty `datasets` — so feel free to open a follow-up issue if you think it's worth tracking, or leave it.

Minor note: `size: "50"` now sits exactly at Azul's hard maximum with no headroom. Since this switches to `fetchAllEntities`, a future tightening on their side would surface as a 400 rather than silent truncation, which is the right failure mode.

